PR #12, opened 2018-11-11
Add "mbed" to keywords

jonas-schievink:
I just searched for "mbed tls" on crates.io and the only result was `mbedtls-sys`. This is clearly also crates.io's fault, but this might be a useful workaround.


jethrogb:
Thanks for the PR! I'll just roll this up in the next release.

PR #13, opened 2018-11-22
Replace expired certificate in tests

ekse:
The certificate used in the tests in expired which causes the tests to fail. This patch replace it with a new certificate, the new expiration date is set to 2046.

I am not able to compile the master branch of this project on my system but I tested this change on my other branch and cherry-picked the commit so it should work.


jethrogb:
Strangely enough, the tests are passing without this PR for me.

jethrogb:
Ah, the `time` feature wasn't being enabled.

jethrogb:
I significantly reduced the changes and fixed the test script.

PR #16, opened 2019-01-17
Expand operations for Mpi

jack-fortanix:
Taken from internal version, `rustfmt`ed, and some adjustments for different feature flags.

### DIFF
--- a/mbedtls/src/lib.rs
+++ b/mbedtls/src/lib.rs
@@ -35,7 +35,6 @@ mod wrapper_macros;
 // ==============
 //      API
 // ==============
-#[allow(dead_code)] // to be exported once the API is more complete
 pub mod bignum;
 mod error;
 pub use error::{Error, Result};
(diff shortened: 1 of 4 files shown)


mbedtls/src/lib.rs
```diff
@@ -36,7 +36,7 @@ mod wrapper_macros;
 //      API
 // ==============
 #[allow(dead_code)] // to be exported once the API is more complete
```
jethrogb: I think this lint should go now

jack-fortanix: Done

jack-fortanix:
A bit of code used  `trim_right_matches`, but this is deprecated starting in 1.33. Since warnings are errors, I changed it to use `trim_end_matches`. But that new API doesn't exist in the old nightly used in the no-`std` tests. What to do?

jethrogb:
Use `trim_right_matches` and apply `#[allow(deprecated)]` in strategic places.

jethrogb:
r=me after you get CI to pass

jack-fortanix:
CI happy now

PR #19, opened 2019-01-21
Add missing error codes

jack-fortanix:


jethrogb:
I'm confused, I thought I had fixed this macro so that rustfmt would pick it up, but it looks like it somehow didn't?

jethrogb:
Can you please nudge the macro/rustfmt a little bit so it gets this enum?

jack-fortanix:
My fault, forgot to `rustfmt` after rebasing

jethrogb:
Great, I'm still confused why it's not formatted on the left of the diff, but whatever.

bors r+

PR #20, opened 2019-01-21
Add functions to get negotiated version and set min/max versions

jack-fortanix:


mbedtls/tests/client_server.rs
```diff
@@ -16,13 +16,14 @@ use mbedtls::rng::CtrDrbg;
...
 use mbedtls::x509::{Certificate, LinkedCertificate, VerifyError};
+use mbedtls::Error;
 use mbedtls::Result as TlsResult;
 
 mod support;
 use support::entropy::entropy_new;
 use support::keys;
 
-fn client(mut conn: TcpStream) -> TlsResult<()> {
+fn client(mut conn: TcpStream, min_minor: i32, max_minor: i32, exp_minor: i32) -> TlsResult<()> {
```
jethrogb: Please use `exp_minor: Result<i32, ()>` instead of -1 to indicate intent

mbedtls/src/ssl/config.rs
```diff
@@ -120,6 +120,14 @@ impl<'c> Config<'c> {
         unsafe { ssl_conf_curves(&mut self.inner, list.as_ptr()) }
     }
 
+    pub fn set_min_version(&mut self, major: i32, minor: i32) {
+        unsafe { ssl_conf_min_version(&mut self.inner, major, minor) };
+    }
+
+    pub fn set_max_version(&mut self, major: i32, minor: i32) {
+        unsafe { ssl_conf_max_version(&mut self.inner, major, minor) };
```
jethrogb: What happens if you pass a value out of range? Like 5, 5.

jack-fortanix: Mbedtls accepts anything, these functions don't do any validation of input just set a var in the `ssl_config`. Appears the client actually offers the max version set here, even though mbedtls doesn't know about TLS 5.5 and if a server ever replied with such a server hello, the client would reject it as something it didn't know. The server looks like it would be ok with a bogus max version since then it would just negotiate whatever the client offered.

We could check that major == 3 and minor <= 3 before calling to save user from a footgun.

mbedtls/tests/client_server.rs
jack-fortanix: Done

mbedtls/src/ssl/config.rs
jack-fortanix: Checks added

jack-fortanix:
bors r=jethrogb

jack-fortanix:
bors r=jethrogb

PR #21, opened 2019-01-21
Add EcPoint, and functions to access key material to Pk

jack-fortanix:


mbedtls/src/pk/mod.rs
```diff
@@ -178,9 +184,124 @@ impl Pk {
...
+                ptr::null_mut(),
+                0,
+                ptr::null_mut(),
+                0,
+                e.as_mut_ptr(),
+                e.len(),
+            )
+            .into_result()?;
+        }
+        Ok(BigEndian::read_u32(&e))
```
jethrogb: [`u32::from_be_bytes`](https://doc.rust-lang.org/std/primitive.u32.html#method.from_be_bytes) is stable now, using that avoids byteorder dependency.

mbedtls/tests/cert.rs
```diff
@@ -0,0 +1,60 @@
+extern crate mbedtls;
```
jethrogb: This and the ECP tests file below make more sense as a unit test, I think.

mbedtls/src/ecp/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+        vector big enough and then resize it down to the actual output
+        length
+
+        ceil(521/8) == 66
+        OS2ECP format is header byte + 2 point elements (or 1 if compressed)
+        so max size is 66*2+1 = 133
+         */
+        let mut olen = 0;
+        let mut buf = Vec::new();
+        buf.resize(133, 0u8);
```
jethrogb: `let mut buf = vec![0u8; 133];`

jack-fortanix: `vec!` is missing from the no-std builds. For some reason it is not exported by alloc crate?

jethrogb: I think you need to add `#[macro_use]` on the `extern crate alloc`

mbedtls/src/pk/mod.rs
jack-fortanix: It is stable, but is not in the old nightly used for the core_io build.

jack-fortanix: Leaving byteorder for now, we can address this when core_io build is updated to a recent nightly. I would take this on, but I am not sure what is needed.

jack-fortanix:
bors r=jethrogb

mbedtls/src/pk/mod.rs
jethrogb: > It is stable, but is not in the old nightly used for the core_io build.

Oh yeah :( never mind then

PR #22, opened 2019-01-22
Change macros so that rustfmt can format them, run rustfmt

jethrogb:
@jack-fortanix sorry this is going to conflict with everything


jack-fortanix:
`rustfmt` is good.

bors r+

PR #23, opened 2019-01-22
Cache cargo

jack-fortanix:
CI build spends most of its time building our dependencies, which are not going to change that often.


jethrogb:
Is it possible to enable this for PR builds but not for bors builds? I'd like to be 100% sure the bors builds build well from scratch.

jack-fortanix:
I'm not sure if that is possible. I'll check later today and if it is, I'll update the PR, or otherwise close.

jethrogb:
Actually, reading up on https://docs.travis-ci.com/user/caching/#rust-cargo-cache, I'm more concerned about `$TRAVIS_BUILD_DIR/target` and less about `~/.cargo`.

jack-fortanix:
I did not realize this caused `target` to be cached, that seems troublesome.

If caching `~/.cargo` is ok we could use 

```
cache:
  directories:
  - $HOME/.cargo
```

plus removing cargo in `staging` to force a clean build for bors.

jethrogb:
Well not caching target means that you're only caching the *download* of the dependencies, not the actual building.

jack-fortanix:
Alas, I had thought (obviously incorrectly) `cargo` cached builds of dependencies in `~/.cargo` so they can be shared among different crates.

Proposed resolution N+1: Cache .cargo and target, `rm -rf` `target` in `staging`, in all cases let the download be serviced from the cached `.cargo` since we should (given a checked in `Cargo.lock`) always be downloading the same bits regardless.

jethrogb:
I second resolution N+1

jack-fortanix:
First build took 13m30s, I restarted and it took 8m30s. That's basically the optimal scenario since the code (and nightly version) did not change, but seems like a win.

PR #25, opened 2019-01-23
Add RFC 6979 ECDSA signatures

jack-fortanix:


jethrogb:
Don't you need to change `mbedtls-sys/build/config.rs`? Add test that `Pk::sign` is non-deterministic.

mbedtls/src/pk/mod.rs
```diff
@@ -401,6 +403,56 @@ impl Pk {
         Ok(ret)
```
jethrogb: Don't you need to change `mbedtls-sys/build/config.rs`? Add test that `Pk::sign` is non-deterministic.

jack-fortanix: You're right thanks

jack-fortanix:
bors r=jethrogb

PR #27, opened 2019-01-24
Change define! macro again

jethrogb:
We used to have assignment in the macro but I couldn't come up with a syntax that would fly with rustfmt when making the change a few days ago. I thought of one now.

Not sure if this is better or not. The advantage is it puts everything on a single line.


jack-fortanix:
Syntax is a little funky but I like having it all on one line.

bors +r

jack-fortanix:
Yes bors r+

jack-fortanix:
Let's try that again

bors r+

PR #31, opened 2019-01-25
Add a feature legacy_protocols to enable/disable SSLv3, TLSv1.0, TLSv1.1

jack-fortanix:
I'd prefer if we could make `legacy_protocols` not in `default`, and instead you only get old TLS if you explicitly ask for it. I'm not sure if that would break anything we care about.


mbedtls/tests/client_server.rs
```diff
@@ -137,17 +139,20 @@ mod test {
             let server_max_ver = config[3];
             let exp_version = config[4];
 
+            if (client_max_ver < 3 || server_max_ver < 3) && !cfg!(feature="legacy_protocols") {
```
jethrogb: It would be nice to test the negative cases, but I guess that would require support on either the client or the server so that doesn't work with conditional compilation.

jethrogb:
> I'd prefer if we could make legacy_protocols not in default, and instead you only get old TLS if you explicitly ask for it. I'm not sure if that would break anything we care about.

I think the current way is fine, it's only enabled by default if you use `mbedtls-sys` directly.

README.md
```diff
@@ -90,6 +91,7 @@ This is a list of the Cargo features available for mbedtls-sys. Features in
 * *havege* Enable the Hardware Volatile Entropy Gathering and Expansion
            (HAVEGE) algorithm.
 * **padlock** Enable support for VIA padlock.
+* **legacy_protocols** Enable support for SSLv3, TLSv1.0 and TLSv1.1
```
jethrogb: Please order these alphabetically (also above)

PR #37, opened 2019-04-03
Add point decompression to EcPoint::from_binary

jack-fortanix:


mbedtls/src/ecp/mod.rs
```diff
@@ -170,18 +177,47 @@ impl EcPoint {
...
+        } else {
+            let mut ret = Self::init();
+            unsafe { ecp_point_read_binary(&group.inner, &mut ret.inner, bin.as_ptr(), bin.len()) }
+                .into_result()?;
+            Ok(ret)
+        }
     }
 
-    pub fn from_components(x: &Mpi, y: &Mpi) -> ::Result<EcPoint> {
+    pub fn from_components(x: Mpi, y: Mpi) -> ::Result<EcPoint> {
```
jethrogb: Public API change: please bump crate version in Cargo.toml

mbedtls/src/bignum/mod.rs
```diff
@@ -237,6 +237,148 @@ impl Mpi {
...
+            let t2 = t.mod_exp(&two, p)?;
+
+            x = (&x * &t)?.modulo(p)?;
+            b = (&b * &t2)?.modulo(p)?;
+            g = t2;
+            r = m;
+        }
+    }
+
+    fn trailing_zeros(&self) -> usize {
```
jethrogb: `pub fn`?

jack-fortanix: Might as well. We didn't need it to be public in the internal crate but this operation comes up a lot in number theoretic algorithms (as can be seen here by it being used in both Jacobi and Shanks-Tonelli functions) and a user of the crate might have need for it.

PR #41, opened 2019-05-03
EC enhancements

jack-fortanix:
Basically a rollup of recent changes to internal version + implicitly declare `handle`/`handle_mut` for all structs via macro

### DIFF
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -11,6 +11,7 @@ use alloc_prelude::*;
 use mbedtls_sys::*;
 
 use core::ptr;
+use core::convert::TryInto;
 use error::IntoResult;
 use private::UnsafeFrom;
 
@@ -26,6 +27,9 @@ use ecp::EcPoint;
 
 #[doc(inline)]
 pub use self::ec::{EcGroupId, ECDSA_MAX_LEN};
+
+#[doc(inline)]
+pub use ecp::EcGroup;
 
 define!(
     #[c_ty(pk_type_t)]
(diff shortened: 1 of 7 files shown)


mbedtls/src/pk/mod.rs
```diff
@@ -139,6 +142,52 @@ impl Pk {
         Ok(ret)
     }
 
+    pub fn generate_ec_using_curve<F: ::rng::Random>(
```
jethrogb: See comment on the internal PR for the API i'd like to see here

mbedtls/src/private.rs
```diff
@@ -26,6 +26,13 @@ pub fn alloc_vec_repeat<F>(mut f: F, data_at_end: bool) -> ::Result<Vec<u8>>
 where
     F: FnMut(*mut c_uchar, size_t) -> c_int,
 {
+    /*
+    Avoid allocating more than a limited amount of memory. In certain conditions
+    with malformed datastructures, mbedtls may return a too-small error regardless of how much
+    buffer space is provided. This causes a loop which terminates with a out of memory panic.
+    */
+    let max_vector_allocation = 4 * 1024 * 1024;
```
jethrogb: `const` makes more sense I think

mbedtls/src/pk/mod.rs
jack-fortanix: @jethrogb Made those changes but I am running into this
```error[E0277]: `?` couldn't convert the error to `error::Error`
   --> mbedtls/src/pk/mod.rs:134:51
    |
134 |             let curve : EcGroup = curve.try_into()?;
    |                                                   ^ the trait `std::convert::From<impl Into<::Error>>` is not implemented for `error::Error`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: consider adding a `where error::Error: std::convert::From<impl Into<::Error>>` bound
    = note: required by `std::convert::From::from`
```
and I'm not sure how to fix it.

mbedtls/src/lib.rs
```diff
@@ -7,6 +7,11 @@
  * according to those terms. */
 
 #![deny(warnings)]
+
+#![allow(stable_features)]
+#![feature(universal_impl_trait)]
+#![feature(try_from)]
```
jethrogb: This is not acceptable. Why is this needed?

jack-fortanix: Because the old nightly used in core_io build requires a feature for try_from https://travis-ci.com/fortanix/rust-mbedtls/jobs/217291163#L1170 and because
```
#![cfg_attr(feature = "core_io", feature(universal_impl_trait))]
#![cfg_attr(feature = "core_io", feature(try_from))] 
```
doesn't work. Obviously feature doesn't work on stable so yeah wtf how to make this work?

jethrogb:
Can you update the nightly that's used along with the core_io crate?

mbedtls/src/lib.rs
jethrogb: Can you update the nightly that's used along with the core_io crate?

jack-fortanix: ok

Goirad:
This is the last one needed to match the internal version, plus ed25519

jethrogb:
@Goirad can you please do the review?

jack-fortanix:
@Goirad fyi #44 is needed also, we use that for Ed25519 in backend

PR #33, opened 2019-03-12
Add support for key wrap as a cipher mode

jack-fortanix:
Change to vendor submitted upstream in PR https://github.com/ARMmbed/mbedtls/pull/2463

PR #43, opened 2019-06-19
Update to mbedtls 2.18.0

### LINKED ISSUE #42: Update to 2.18.0

jack-fortanix:
Closes #42


jethrogb:
Hmm I think we should probably follow suit and split into two crates

jack-fortanix:
I'm not sure if a split makes sense, because mbedtls still requires an embedded copy of mbed-crypto to build, and rexports mbed-crypto as part of the mbedtls API. So if we split into two crates, we end up with two copies of mbed-crypto in the resulting binary.

jethrogb:
The embedded copy is used for more than just include files?

jack-fortanix:
Yes it builds the version in `crypto/lib` to form `libmbedcrypto`. The crypto code in `mbedtls` itself is dead - if you check the `CMakeLists.txt` in `library` you see that `src_crypto` var is still set, but is just ignored now. No idea why they haven't removed that code entirely.

jethrogb:
Ok let's keep it as a single crate for now, we can reevaluate later.

jack-fortanix:
@jethrogb ping

jethrogb:
I think you still need to bring over a bunch of changes we did internally.

jack-fortanix:
Such as? Everything I see in the internal log is covered by #33 #41 #44

jethrogb:
I had a bunch of comments on your internal PR, those don't seem to have been fixed here. e.g. `mbedtls-sys/build/config.rs`

jack-fortanix:
Updated config.rs. No changes to headers.rs needed

mbedtls/src/bignum/mod.rs
```diff
@@ -172,8 +172,9 @@ impl Mpi {
 
         let s = String::from_utf8(buf).expect("from_utf8 can't fail on radix-N data");
 
-        #[allow(deprecated)]
-        Ok(s.trim_right_matches(char::from(0)).to_owned())
+        // There is a null terminator plus (possibly) some garbage data. Remove it
+        let s : Vec<&str> = s.split(|c| c == char::from(0)).collect();
+        Ok(s[0].to_owned())
```
jethrogb: We fixed this internally as well

jethrogb:
Can you update the nightly that's used?

PR #44, opened 2019-06-26
Support custom PK algo types

jack-fortanix:
Same as internal version

### DIFF
--- a/mbedtls/src/ssl/ciphersuites.rs
+++ b/mbedtls/src/ssl/ciphersuites.rs
@@ -7,6 +7,7 @@
  * according to those terms. */
 
 use mbedtls_sys::types::raw_types::c_int;
+use mbedtls_sys::*;
 
 define!(
     #[non_exhaustive]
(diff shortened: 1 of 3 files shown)


mbedtls/src/ssl/ciphersuites.rs
```diff
@@ -7,6 +7,7 @@
  * according to those terms. */
 
 use mbedtls_sys::types::raw_types::c_int;
+use mbedtls_sys::*;
```
Goirad: Why is this import here if nothing is being added to the file?

Goirad:
Also it would be nice to have one more PR before doing a release that just updates everything for 2018 edition and resolves deprecation warnings

mbedtls/src/ssl/ciphersuites.rs
jack-fortanix: I had to change `define_enum` macro to not assume the values are in `mbedtls_sys::` since we need to reference a value that is not defined in that crate (`CUSTOM_PK_TYPE`)

Goirad: I see

Goirad:
Oh it looks like there are some conflicts

jack-fortanix:
@Goirad I will create a PR updating edition

PR #47, opened 2019-07-30
Upgrade to 2018 edition

jack-fortanix:
This only updates `mbedtls` and not `mbedtls-sys` because the bindings get generated to use `::types::...` when they should use `crate::types::...` in 2018. I couldn't figure out a way to fix this. It probably requires upgrading bindgen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,7 +206,7 @@ dependencies = [
  "cc 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
  "core_io 0.1.20190701 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
- "mbedtls-sys-auto 2.16.0",
+ "mbedtls-sys-auto 2.18.0",
  "rand 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "rc2 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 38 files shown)


mbedtls-sys/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "2.18.0"
 authors = ["Jethro Beekman <jethro@fortanix.com>"]
 build = "build/build.rs"
 license = "Apache-2.0/GPL-2.0+"
+#edition = "2018"
```
Goirad: Can this be uncommented now?

jack-fortanix: No due to issue with bindgen output mentioned in summary

Goirad: Oh right sorry, thoguh this was in mbedtls no mbedtls-sys

Cargo.lock
```diff
@@ -206,7 +206,7 @@ dependencies = [
  "cc 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
  "core_io 0.1.20190701 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
- "mbedtls-sys-auto 2.16.0",
+ "mbedtls-sys-auto 2.18.0",
```
jethrogb: How come this is only changing now?

jack-fortanix: No good reason. Was missed in 2.18 upgrade

PR #49, opened 2019-08-14
Add Version enum for TLS

jack-fortanix:
Adapted from ced45e6c2e of internal version

### DIFF
--- a/mbedtls/tests/client_server.rs
+++ b/mbedtls/tests/client_server.rs
@@ -29,7 +29,7 @@ fn client(
     mut conn: TcpStream,
     min_version: Version,
     max_version: Version,
-    exp_version: Result<Version, ()>) -> TlsResult<()> {
+    exp_version: Option<Version>) -> TlsResult<()> {
     let mut entropy = entropy_new();
     let mut rng = CtrDrbg::new(&mut entropy, None)?;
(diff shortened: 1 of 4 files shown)


mbedtls/tests/client_server.rs
```diff
@@ -81,29 +80,29 @@ fn client(
 
 fn server(
     mut conn: TcpStream,
-    min_minor: i32,
-    max_minor: i32,
-    exp_minor: Result<i32, ()>,
+    min_version: Version,
+    max_version: Version,
+    exp_version: Result<Version, ()>,
 ) -> TlsResult<()> {
```
Goirad: Why is the expected version passed as a result? It also seems to me like it is acting semantically as an `Option<Version>`. I see that it is like this since before this particular PR, but I'd still like to know.

jack-fortanix: It is `None` if (given the client+server configurations) it is expected to be impossible to negotiate. F.ex. if client is TLS 1.0 only and server is 1.2 only, then the expected version of the session is not any of the values enumerated in `Version`.

jack-fortanix: Sorry I misparsed your question. It should be an `Option`.

Goirad: Ah ok I understand

PR #50, opened 2019-08-21
Implement cert expiry checking without libc

Goirad:
Add features and implementations of time related functions to use when `libc` isn't available.
Because of feature weirdness I added a `use_libc` feature for those that want to just use libc.
The `gmtime_r` function is written with chrono, as is the custom `time` function.


jack-fortanix:
Changes all look good. A test would be nice (eg using a certificate with known validity range), as would checking the build for some of the relevant configurations in `ct.sh` - right now we are not testing time+no-std builds so I don't think this code is being built at all.

mbedtls/tests/client_server.rs
```diff
@@ -56,10 +58,11 @@ fn client(
                 s
             }
             Err(e) => {
-                assert!(exp_version.is_none());
```
jack-fortanix: Why this change?

mbedtls/tests/client_server.rs
```diff
@@ -102,11 +105,12 @@ fn server(
             s
         }
         Err(e) => {
-            assert!(exp_version.is_none());
+            //assert!(exp_version.is_none());
```
jack-fortanix: Also ?

PR #52, opened 2019-08-22
Add bytes_available on Session

jack-fortanix:
rust-native-tls wants this


mbedtls/src/ssl/context.rs
```diff
@@ -225,6 +225,10 @@ impl<'a> Session<'a> {
         self.inner.major_ver
     }
```
Goirad: Can you add some documentation?

PR #53, opened 2019-08-22
impl Send/Sync for Certificate

jack-fortanix:
AFAICT this is safe. And needed for native-tls.


jack-fortanix:
#70 does what is needed

PR #54, opened 2019-08-28
update bitflags and remove mem::uninitialized

Goirad:
As the title says, solves deprecation warning associated with those two issues. Probably worth making an issue or otherwise noting that a better strategy around `MaybeUninit` is required.

This is also required to get CI passing again


Goirad:
I have also removed the cargo cache line from the travis.yml file, since that is seemingly cause builds to hand and time out

PR #55, opened 2019-09-04
If feature=force_aesni_support then disable the AES T-table code

jack-fortanix:
This prevents any possibility of calling side-channel insecure code, and also reduces code size.


mbedtls-sys/Cargo.toml
```diff
@@ -36,6 +36,7 @@ default = ["std", "zlib", "time", "pthread", "aesni", "padlock", "legacy_protoco
 std = []
 custom_printf = []
 custom_has_support = []
+force_aesni_support = []
```
jethrogb: The feature in this create should just be called `aes_alt`

jethrogb:
Do we test force_aesni_support in CI right now? If so r=me. Otherwise, please add that and then r=me.

jack-fortanix:
Build was tested (via sgx target) but was not running the tests. Added.

PR #57, opened 2019-09-16
Add accessors for X.509 certificate fields

jack-fortanix:
This changes the behavior of `issuer_raw` and `subject_raw`. Currently those return the same value as issuer/subject, except as `Vec<u8>` instead of `String`s, which is just duplicating `String::into_bytes`. IMO they should instead return the raw binary DER which allows accessing fields in the DN which mbedtls doesn't know about or can't handle: `mbedtls_x509_dn_gets` prints OIDs it doesn't know as "??", and characters in ranges 0...32 and 128..160 as '?', meaning it mangles UTF-8 and UCS-2 names.

### DIFF
--- a/mbedtls/src/x509/certificate.rs
+++ b/mbedtls/src/x509/certificate.rs
@@ -23,6 +23,13 @@ use crate::private::UnsafeFrom;
 use crate::rng::Random;
 use crate::hash::Type as MdType;
 
+#[derive(Debug,Copy,Clone,Eq,PartialEq)]
+pub enum CertificateVersion {
+    V1,
+    V2,
+    V3
+}
+
 define!(
     #[c_ty(x509_crt)]
     struct Certificate;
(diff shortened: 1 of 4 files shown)


mbedtls/src/x509/certificate.rs
```diff
@@ -180,6 +189,26 @@ impl LinkedCertificate {
         unsafe { ::core::slice::from_raw_parts(self.inner.raw.p, self.inner.raw.len) }
     }
```
Goirad: Might this be better as an enum?

mbedtls/src/x509/certificate.rs
```diff
@@ -180,6 +189,26 @@ impl LinkedCertificate {
...
 
+    pub fn version(&self) -> i32 {
+        self.inner.version
+    }
+
```
Goirad: Might this be better as an enum?

jack-fortanix: enum seems fine and is probably more convenient, will change

PR #58, opened 2019-09-18
Add accessor to get the sign of a Mpi

jack-fortanix:


mbedtls/src/bignum/mod.rs
```diff
@@ -140,6 +146,14 @@ impl Mpi {
         Ok(self.get_limb(0) as u32)
     }
 
+    pub fn sign(&self) -> Sign {
+        if self.inner.s < 0 {
+            Sign::Negative
+        } else {
+            Sign::Positive
+        }
```
Goirad: Is it intentional for 0 to be positive? The mathematician in me says that ideally there should be three enums, negative positive and zero, or least negative and nonegative, but I can understand if this just convention in this particular field or application.

jack-fortanix: In sign-magnitude representation bignums, normally zero is considered positive, because that ends up being convenient for implementation reasons. But I agree it is mathematically speaking not so great. To detect zero we have to add an explicit comparison if s == 1. I'm ok with that. For what I need I only care about < 0 vs >= 0 (because ASN.1 uses a different encoding for negative).

jack-fortanix: Add zero

Goirad: https://tls.mbed.org/api/bignum_8h.html seems to lack any documentation or methods related to the implementation's notion of sign, at least not besides `integer sign` for the `s` field of an mpi which is unfortunate

jack-fortanix: Yeah my interpretation is just based on inspection of library/bignum.c since I myself was not sure if they used s==0 to indicate a zero Mpi

PR #63, opened 2019-10-08
WIP: full no_std support

### LINKED ISSUE #56: Update to mbedtls 2.19.1
Apparently released on Sep 6 - https://github.com/ARMmbed/mbedtls/blob/development/ChangeLog

Fixes a couple of side channels (though we already avoid the one affecting deterministic ECDSA), a faster X25519 for x86, and adds mbedtls_ssl_session_save/mbedtls_ssl_session_load which will be very useful when implementing support for cross-node ticket keys.

### LINKED ISSUE #62: no_std compilation
I'm having alot of trouble building this crate on a truly no_std platform (thumbv7m).

```
cargo build --no-default-features --features core_io --target=thumbv7m-none-eabi -j1
```

Initial findings are that `yasna` uses std, but beyond that there seems to be a whole heap of other issues like dev dependencies attempting to build and failing for that target (which I think it actually a cargo issue).

MabezDev:
Closes #62 

### Unresolved issues:

* ~~What to do about yasna? AFAIK it does not offer a std feature that can be disabled~~
* What to do about the dev dependancy issues explored in #62

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,4 +32,5 @@ script:
   - rustup toolchain add $CORE_IO_NIGHTLY
   - rustup toolchain add nightly
   - rustup target add --toolchain nightly x86_64-fortanix-unknown-sgx
+  - rustup target add --toolchain nightly thumbv7m-none-eabi
   - ./ct.sh
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,11 +154,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fuchsia-cprng"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "generic-array"
 version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -169,11 +164,6 @@ dependencies = [
 [[package]]
 name = "glob"
 version = "0.2.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
-name = "hex"
-version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -232,17 +222,12 @@ dependencies = [
  "cc 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "core_io 0.1.20190701 (registry+https://github.com/rust-lang/crates.io-index)",
- "hex 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
  "mbedtls-sys-auto 2.18.0",
- "rand 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "rc2 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rs-libc 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.97 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_cbor 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.97 (registry+https://github.com/rust-lang/crates.io-index)",
  "spin 0.4.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "yasna 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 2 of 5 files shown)


jethrogb:
Thanks for the PR! You probably need to update `.travis.yml` to install the target as well.

MabezDev:
Yeah, adding the target helps :). I've cherry picked these changes to the parent commit of the yasna merge, which almost compiles but I'm greeted with this:

```rust
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> mbedtls/src/cipher/raw/serde.rs:301:5
    |
301 |     ::core::mem::transmute(ctx)
    |     ^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: source type: `mbedtls_sys::cipher_context_t` (544 bits)
    = note: target type: `[u8; 96]` (768 bits)

error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> mbedtls/src/cipher/raw/serde.rs:305:5
    |
305 |     ::core::mem::transmute(ctx)
    |     ^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: source type: `mbedtls_sys::aes_context` (2240 bits)
    = note: target type: `[u8; 288]` (2304 bits)

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0512`.
error: Could not compile `mbedtls`.
```
I can comment this out to make sure it does actually compile but that's not a good idea in the long run.

As far as I can see, the mbed source is tracked inside mbed-sys-auto so I'm guessing this isn't due to changes in the C api, and more likely and issue with compiling with an arm compiler. Any thoughts?

You can try it out in [this branch](https://github.com/MabezDev/rust-mbedtls/tree/no-std-cherry-picked).

jack-fortanix:
Likely a combination of structure padding and also different pointer sizes. @jethrogb these sizes are checked because the cipher_context serialization code uses them, right? Can we switch serialization to using a generic format that doesn't rely on the layout of cipher_context_t? IIRC this was discussed on the RTE slack a while back.

jethrogb:
Yeah RH did ask about that

jethrogb:
@MabezDev we just merged a couple of PRs that should make getting this to work easier

MabezDev:
Thanks, that helps alot! From master I now only need to change byteorder set use default-features = false for it compile. 

Unfortunately using it in a project I run into https://github.com/rust-lang/cargo/issues/5730 . There seem to be a few work around, none of them particularly nice. 

* One being https://github.com/auxoncorp/cargo-5730. 
* The second solution I've seen it by forcing the user to have bindgen installed in their $PATH, hence removing the build dependencies all together and invoking bindgen manually. 
* Another 'solution' I've seen is forking all std enabled crates, disabling std and using cargo patch - but I don't think that would work here as std is required and quite frankly that is an insane solution.

With no clear fix to this cargo issue, would you be willing to accept one of these work arounds?

jethrogb:
I kind of like the solution of https://github.com/auxoncorp/cargo-5730, but I think it's too complicated. I came up with the following:

Make all build dependencies optional, add a feature to enable then, and allow running the build script via `cargo run`:
```toml
# Cargo.toml
[package]
build = "build.rs"

[[bin]]
name = "build-script-build"
path = "build.rs"

[features]
build = ["dep"]

[build-dependencies]
dep = { ..., optional = true }
```

In the build script, if the feature isn't enabled, re-execute the build script:
```rust
// build.rs
#[cfg(not(feature = "build"))]
fn main() {
    let cargo = std::env::var_os("CARGO").unwrap()
    let args = [
        /* cargo, */ "run",
        "--bin", "build-script-build",
        "--features", "build",
        "--target-dir", /* $OUT_DIR/bindgen */
    ];
    let mut target_dir = std::path::PathBuf::from(std::env::var_os("OUT_DIR").unwrap());
    target_dir.push("bindgen");
    assert!(std::process::Command::new(&cargo).args(args).arg(&target_dir).status().unwrap().success())
}

#[cfg(feature = "build")]
fn main() {
    // normal build script
    // TODO: do nothing if this is invoked as a build script
}
```

Note, during a normal build, the build script gets executed 3 times: once as the original build script, then again as the build script with the feature enabled, then as the `bin` target for `cargo run`. Some logic should be added so the second execution does nothing.

MabezDev:
That technique seems to work quite well, though I had to specify the optional build dependencies inside the normal dependacies too, but I think there is a way to avoid this. 

Very close to being able to build code for arm, now failing at the linking stage with:
```
 = note: /usr/lib/gcc/arm-none-eabi/9.2.0/../../../../arm-none-eabi/bin/ld: error: no memory region specified for loadable section `.got'
          collect2: error: ld returned 1 exit status
          

error: aborting due to previous error
```

the got section indicates dynamic relocation (see: https://github.com/rust-embedded/cortex-m-rt/blob/master/link.x.in#L143), I'm guessing something isn't being built statically - just haven't figured out what yet.

MabezDev:
Well that took longer than expected. Turned out cmake was injecting the `-fPIC` flag which allowed the compiler to create relocatable sections; cmake-rs currently doesn't support disabling PIC but a [PR exists](https://github.com/alexcrichton/cmake-rs/pull/89). I now have built mbedtls withing an thumbv7 project and have it running (yay!), though not actually doing anything with it yet.

I'll spend a bit of time cleaning up the changes I've made and submit them hopefully tomorrow. I'll probably close this and open a fresh PR when I do.

MabezDev:
Closed in favour of #67

PR #65, opened 2019-10-08
Only require chrono when we actually use it

jack-fortanix:
See #62

### DIFF
--- a/mbedtls/Cargo.toml
+++ b/mbedtls/Cargo.toml
@@ -19,7 +19,7 @@ keywords = ["MbedTLS","mbed","TLS","SSL","cryptography"]
 
 [dependencies]
 bitflags = "1"
-chrono = "0.4"
+chrono = { version = "0.4", optional = true }
 core_io = { version = "0.1", features = ["collections"], optional = true }
 spin = { version = "0.4.0", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


Goirad:
@jack-fortanix This looks good, definitely an oversight on my part. It looks like we just need to update syntex-syntax since NLL error suppression has finally been disabled

Goirad:
On second look though it seems that we would have to update bindgen to update syntex-syntax, so this became a lot more complicated

jack-fortanix:
Yeah I was just looking at that. I think the thing to do is for now is pin the SGX build to use nightly-2019-07-11

Goirad:
Don't we use bindgen for all targets?

jack-fortanix:
Yes, but the problem with syntex-syntax is only on very recent nightly. On stable/beta/older nightly, it warns.

This certainly puts a fire under #5 since this means in a few months we won't have a working build on stable anymore. But pinning the nightly build back a ways (and specifically to what we are using in roche) lets things carry on for now. And re SGX support, we really care about that one snapshot vs whatever nightly happens to be doing this week, so I think it makes more sense to pin in any case.

Goirad:
Oh ok I understand, that makes sense

PR #66, opened 2019-10-08
Split up Travis into 4 jobs and pin to a specific nightly for SGX

jack-fortanix:
This improves parallelism (Travis will run 5 jobs at once) and avoids redundancy (the core_io + nightly builds are being run twice, in the stable and beta builds).

Pins the SGX build to 2019-07-11 as with most recently nightly, syntex_syntax fails to compile. Fixing that requires upgrading bindgen (#5)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ branches:
     # Not really necessary, just to get a green badge on “master”
     - master
 language: rust
+os: linux
 dist: xenial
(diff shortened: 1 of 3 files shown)


jethrogb:
> Pins the SGX build to 2019-07-11 as with most recently nightly, syntex_syntax fails to compile.
What's the error?

PR #67, opened 2019-10-11
no-std-build-fixes:

### LINKED ISSUE #62: no_std compilation

MabezDev:
Closes #62

* Removes dev/build dependencies, and instead are listed as optional
normal deps (See: https://github.com/fortanix/rust-mbedtls/pull/63#issuecomment-539974309)
* Adds the abort feature to disable unwinding
* Defaults to not passing `-fPIC` to the C compiler
* Adds the cross-compile feature to prevent building for the host and
target platforms for cross compile targets
* Adds the mbedtls-test feature to enable the dev-deps for tests and
examples

### Unresolved
* How to make the sanity checks in `mbedtls/src/cipher/raw/serde.rs` platform independant
* The build script work around creates a cargo warning
* Relies on git version of cmake-rs currently
* Build deps have to been declared as normal optional deps too, even though the dependacies are never actually built

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -120,7 +120,7 @@ dependencies = [
 [[package]]
 name = "cmake"
 version = "0.1.42"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/lthiery/cmake-rs#d7edd3b22c745a9186bacffab2e280f7adb6bc1a"
 dependencies = [
  "cc 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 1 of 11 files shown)


jethrogb:
> Build deps have to been declared as normal optional deps too, even though the dependacies are never actually built

Hmm yeah. This is because we're trying to use `cargo run`. It would be nice if cargo had a direct way to only execute the build script. Maybe instead, the cargo run execution should just find out where the build script output is (cargo writes this to disk) and echo that instead of re-running. That would then also get around the cargo warning.

MabezDev:
Hmm I've had I look around, I'm sure there is a good way of finding the build script executable location. Another thing I tried was doing all the building inside a difference builder binary, then invoking cargo run on that inside the build.rs. Unfortunately this just loops forever because the build.rs runs when you try and run the builder binary.

MabezDev:
Time flies :D. With cargo's `resolver = 2` I think most of the issues here are solved. I may PR some of the changes here if/when I need this crate. Regardless, thanks for the help so far.

PR #70, opened 2019-10-15
Don't parse Certificates in Pkcs12 until requested

jack-fortanix:
This avoids having any C structures in Pkcs12 which makes life easier in rust-native-tls land

Also makes parsing of structures more tolerant, previously if certificate parsing failed the whole structure parse was abandoned. Now we just ignore certs we cannot parse.

### DIFF
--- a/mbedtls/src/pkcs12/mod.rs
+++ b/mbedtls/src/pkcs12/mod.rs
@@ -918,6 +918,9 @@ mod tests {
 
         let certs = pfx.certificates().collect::<Vec<_>>();
         assert_eq!(certs.len(), 1);
+        for cert in certs {
+            assert!(cert.0.is_ok());
+        }
 
         let keys = pfx.private_keys().collect::<Vec<_>>();
         assert_eq!(keys.len(), 1);
@@ -953,6 +956,9 @@ mod tests {
 
         let certs = pfx.certificates().collect::<Vec<_>>();
         assert_eq!(certs.len(), 2);
+        for cert in certs {
+            assert!(cert.0.is_ok());
+        }
 
         let keys = pfx.private_keys().collect::<Vec<_>>();
         assert_eq!(keys.len(), 1);
@@ -974,6 +980,9 @@ mod tests {
 
         let certs = parsed_pfx.certificates().collect::<Vec<_>>();
         assert_eq!(certs.len(), 1);
+        for cert in certs {
+            assert!(cert.0.is_ok());
+        }
 
         let keys = parsed_pfx.private_keys().collect::<Vec<_>>();
         assert_eq!(keys.len(), 1);
@@ -997,6 +1006,9 @@ mod tests {
 
         let certs = pfx.certificates().collect::<Vec<_>>();
         assert_eq!(certs.len(), 1);
+        for cert in certs {
+            assert!(cert.0.is_ok());
+        }
 
         let keys = pfx.private_keys().collect::<Vec<_>>();
         assert_eq!(keys.len(), 1);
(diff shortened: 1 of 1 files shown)


mbedtls/src/pkcs12/mod.rs
```diff
@@ -843,7 +841,9 @@ impl Pfx {
                 for sb in &d.0 {
                     if let &Pkcs12BagSet::Cert(ref cb) = &sb.bag_value {
                         if let Some(ref cert) = cb.cert {
-                            certificates.push((cert.clone(), sb.friendly_name()));
+                            if let Ok(cert) = Certificate::from_der(cert) {
```
jethrogb: There should probably be some indication to the caller that there are more certificates than returned.

jack-fortanix: Maybe change return value to `Pkcs12Result<Vec<(Result<Certificate, MbedtlsError>, Vec<String>)>>` so if a parse error occurs the return from mbedtls is visible to caller?

jethrogb: How about we make this an iterator instead?

jack-fortanix: Sounds good

jack-fortanix: I rewrote the function in a functional style which produces an iterator, but I don't see any good way of returning it. Of the options suggested in https://stackoverflow.com/questions/27535289/what-is-the-correct-way-to-return-an-iterator-or-any-other-trait the only one that really works is Box<> which I'm not sure is desirable.

jethrogb: I was thinking more along these lines:
```rust

    fn authsafe_decrypted_contents(&self) -> impl Iterator<Item=&SafeBag> {
        self.authsafe.contents.iter()
            .flat_map(|d| if let AuthenticatedSafe::Data(ref d) = d { d.0.as_slice() } else { &[] })
    }

    /// Return the certificates stored in this Pfx along with a possibly empty list
    /// of "friendly names" which are associated with said certificate.
    /// Some or all of the certificates stored in a Pfx may be encrypted in which case
    /// decrypt must be called to access them.
    pub fn certificates<'a>(&'a self) -> impl Iterator<Item=(Result<Certificate, crate::Error>, Vec<String>)> + 'a {
        self.authsafe_decrypted_contents()
            .filter_map(|sb| if let Pkcs12BagSet::Cert(CertBag(Some(cert))) = &sb.bag_value {
                Some((Certificate::from_der(cert), sb.friendly_name()))
            } else { None })
    }

    /// Return the private keys stored in this Pfx along with a possibly empty list
    /// of "friendly names" which are associated with said private key.
    pub fn private_keys<'a>(&'a self) -> impl Iterator<Item=(Result<Pk, crate::Error>, Vec<String>)> + 'a {
        self.authsafe_decrypted_contents()
            .filter_map(|sb| 
                match &sb.bag_value {
                    Pkcs12BagSet::Pkcs8(pkcs8) | Pkcs12BagSet::Key(KeyBag { pkcs8 }) => 
                        Some((Pk::from_private_key(pkcs8, None), sb.friendly_name())),
                    _ => /* not a private key */ None
                }
            )
    }
```

mbedtls/src/pkcs12/mod.rs
```diff
@@ -932,25 +907,27 @@ mod tests {
         let parsed_pfx = Pfx::parse(pfx_bits).unwrap();
 
         // Certs are not encrypted in this Pfx
-        let certs = parsed_pfx.certificates().unwrap();
```
jethrogb: For all tests: Should probably check the results inside the vec as well

PR #72, opened 2019-10-25
update to bindgen 0.51.0

haraldh:
This patch lets us compile mbedtls-sys with recent clang versions.

~~It still has a hack calling "sed" in it, which fixes some of the enum
generated names, but it is a good starting point.~~


haraldh:
Got rid of the `sed`

jack-fortanix:
LGTM. @jethrogb ?

haraldh:
@jack-fortanix  @jethrogb 
friendly ping

jethrogb:
Final PR: #152

PR #73, opened 2019-10-25
unfreeze nightly ci after syntex-syntax update

Goirad:


.travis.yml
```diff
@@ -19,6 +19,7 @@ addons:
 rust:
    - stable
    - beta
+   - nightly
    - nightly-2019-07-01
    - nightly-2019-07-11
```
jethrogb: Shouldn't one of these have been removed?

Goirad: The intention is to update 07-11 to whatever the next nightly we end up using for our toolchain is

PR #77, opened 2019-11-04
Update bindgen to 0.19.2

Goirad:
This change makes it so that update also applies to crates that rely on
mbedtls.


jethrogb:
This is not necessary. Bindgen 0.19.0 works fine. Please revert.

PR #82, opened 2019-11-12
Silence unsafe FFI warnings

Goirad:
The rust compiler is issuing false positive improper_ctypes warning, but
this is not an issue because only pointers are being passed to mbedtls,
and they are used completely opaquely.


jethrogb:
This is fine, but ideally this would be reported upstream so they fix the lint.

jethrogb:
https://github.com/rust-lang/rust/pull/65134#issuecomment-553327135

PR #83, opened 2019-11-12
Bump SGX toolchain to match what we are using in backend

jack-fortanix:


jack-fortanix:
I forgot to also update the version of rustc that will be installed, which had the effect of disabling the SGX build. Added a check in ct.sh so that can't happen again.

PR #87, opened 2019-12-09
Version bump for release

Goirad:


mbedtls/Cargo.toml
```diff
@@ -33,7 +33,7 @@ rc2 = { version = "0.3", optional = true }
 rs-libc = "0.1.0"
 
 [dependencies.mbedtls-sys-auto]
-version = "2.18.0"
```
jethrogb: Unnecessary bump

Goirad: But the change we want to pick up is in mbedtls-sys-auto isn't it?

Goirad: Or do you mean that it goes from `2.18.0` to `2.18.2`?

jethrogb: If 2.18.0 is working fine for people then there's no reason to force them to upgrade.

jethrogb: So this means this file doesn't need to change at all.

Goirad: Understood

Issue #5, opened 2018-10-30
Upgrade bindgen


jethrogb:
This crate is currently using an old version of bindgen which requires LLVM 3.8 and no newer. The API and behavior has changed significantly since then, so it'll be a bit of work to upgrade.

ekse:
I would like to give this a try. Since Builder::remove_prefix was removed, are you ok with using names with the "mbedtls_" prefix, except when we reexport values?

jethrogb:
Oh I forgot about this issue. See https://github.com/rust-lang-nursery/rust-bindgen/issues/428 . I'd really like to automate this since I'm not a big fan of having to use the prefixes everywhere.

ekse:
I worked on this issue in the last few days, here is a status update on my progress so far. I would like to discuss potential approaches to deal with issues I encountered. You can find my work branch at https://github.com/ekse/rust-mbedtls/tree/bindgen-upgrade.

First, a quick note on why I want to upgrade. The current master branch does not compile on my computers, it fails with the error `Bindgen ERROR: unsupported type CXTypeKind(119)`. Upgrading bindgen fixes this issue. I get this build error on libclang 3.9, 4.0 and 6.0.

### "mbedtls_" prefix
To deal with the "mbedtls_" prefix, I tried the approach suggested by the bindgen maintainer to rename the imports with `use x as y` statements. To reduce the amount of boilerplate,  I experimented with creating a procedural macro "mbedtls_use" https://github.com/ekse/rust-mbedtls/tree/bindgen-upgrade/mbedtls-use. An example use of this macro looks like this: (https://github.com/ekse/rust-mbedtls/blob/bindgen-upgrade/mbedtls/src/cipher/raw/serde.rs#L24)

```rust
#[mbedtls_use]
use {
    mbedtls_aes_context, mbedtls_cipher_context_t, mbedtls_cipher_id_t, mbedtls_cipher_mode_t,
    mbedtls_des3_context, mbedtls_des_context, MBEDTLS_CIPHER_ID_3DES, MBEDTLS_CIPHER_ID_AES,
    MBEDTLS_CIPHER_ID_DES, MBEDTLS_MODE_CBC, MBEDTLS_MODE_CFB, MBEDTLS_MODE_CTR,
};
```

[...]

PR #88, opened 2019-12-23
Update bindgen [WIP]

jseyfried:
Closes https://github.com/fortanix/rust-mbedtls/issues/5.

Also bump `serde_derive` minor version to match that of `serde` and to reduce Cargo.lock bloat associated with this PR.


jethrogb:
Final PR: #152

jethrogb:
Fixed in #152

PR #90, opened 2020-01-20
Fix ECDSA side channel

jack-fortanix:
Backport of https://github.com/ARMmbed/mbed-crypto/commit/247c4d3c8876dbf52c6c100600ea9b9840cbc904
which addresses the attack described in https://eprint.iacr.org/2020/055.pdf


jack-fortanix:
@jethrogb ping

PR #91, opened 2020-01-22
Parse RSA CRT parameters from PKCS1 private keys

jack-fortanix:
Currently they are ignored in the serialized key and then regenerated from P/Q/D. But that exposes key loading to a side channel attack on the modular inversion and GCD bignum functions when QP is computed. [And probably similar issues for DP/DQ during the division step but no attack there has been published yet.]

Also this reduces computational overhead of loading RSA private keys from memory which will be nice for us in roche.

Backport of https://github.com/ARMmbed/mbed-crypto/pull/352


jack-fortanix:
@jethrogb ping

PR #92, opened 2020-02-26
Added support for SSE2 and enabled it by default. 

raoulstrackx:
SSE2 is used in bignum.c and should make RSA operations much faster.


mbedtls-sys/Cargo.toml
```diff
@@ -32,7 +32,7 @@ cmake = "0.1.17"
 # * strstr/strlen/strncpy/strncmp/strcmp/snprintf
 # * memmove/memcpy/memcmp/memset
 # * rand/printf (used only for self tests. optionally use custom_printf)
-default = ["std", "zlib", "time", "pthread", "aesni", "padlock", "legacy_protocols"]
+default = ["std", "zlib", "time", "pthread", "aesni", "padlock", "legacy_protocols", "sse2"]
```
jack-fortanix: Does having "sse2" as part of default have implications for non-x86 builds?

jack-fortanix:
@raoulstrackx #96 implies we at the least don't want this feature used in SGX, right?

raoulstrackx:
Sorry this was an oversight by me. SSE2 is only 64-bit and thus useless for most people. Will close the PR.

PR #94, opened 2020-03-09
Fix docs on macro invocations

jethrogb:


jethrogb:
Self-reviewing to get the build out

bors r+

PR #96, opened 2020-03-10
Adding mpi_force_c_code feature

raoulstrackx:
Adding C code for mpi so that LVI passes can be optimized.

### DIFF
--- a/mbedtls-sys/vendor/include/mbedtls/bn_mul.h
+++ b/mbedtls-sys/vendor/include/mbedtls/bn_mul.h
@@ -46,7 +46,7 @@
 
 #include "bignum.h"
 
-#if defined(MBEDTLS_HAVE_ASM)
+#if defined(MBEDTLS_HAVE_ASM) && !defined(MBEDTLS_MPI_FORCE_C_CODE)
 
 #ifndef asm
 #define asm __asm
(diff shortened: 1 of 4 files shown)


mbedtls-sys/vendor/include/mbedtls/bn_mul.h
```diff
@@ -195,6 +196,25 @@
         : "b" (b)                           \
         : "rax", "rdx", "r8"                \
     );
+#else
+// Code copied from git https://github.com/ARMmbed/mbedtls.git commit 5f80040729cfbe0cf61a17cbceb64375d4b0acd3 crypto/include/mbedtls/bn_mul.h
```
jethrogb: Why copy? The code is already here at the end of the file.

mbedtls/Cargo.toml
```diff
@@ -57,6 +57,7 @@ spin_threading = ["threading","spin","mbedtls-sys-auto/custom_threading"]
 sgx = ["std", "rust_threading", "rdrand", "force_aesni_support"]
```
jack-fortanix: Should mpi_force_c_code be added to sgx feature?

mbedtls-sys/vendor/include/mbedtls/bn_mul.h
raoulstrackx: Ok resolved, only changing preprocessor conditionals

jethrogb:
Please bump patch versions in Cargo.toml

mbedtls/Cargo.toml
raoulstrackx: I won't add it here as part of sgx. When we start using machines with hardware mitigations against LVI, we won't be able to get rid of the option.

PR #97, opened 2020-03-20
Add ability to update the state of a private key

jack-fortanix:
Needed for stateful signature schemes like LMS or XMSS


jack-fortanix:
This ended up not being necessary

PR #99, opened 2020-03-24
Support upstream Mbed TLS

gbryant-arm:
AFAIK, in the current state of this project, the Mbed TLS "vendor" (hardcoded) version is tightly coupled to the wrapper and difficult to upgrade to a more recent version <del>since Mbed TLS is now built around a self-contained crypto module.</del>
**Mbed Crypto is actually back into Mbed TLS** (https://github.com/ARMmbed/mbed-crypto/issues/374).
This work aims at supporting upstream Mbed TLS and potentially decoupling it from the wrapper.

What has been done so far:
* Removed the hardcoded version of Mbed TLS
* Added upstream Mbed TLS as a submodule
* Added the crypto module to the build system
  + A new header dedicated to the crypto module is generated and passed to bindgen
* Adjusted the rest of the wrapper to the new version of Mbed TLS
* Removed the crypto module from the build system (revert)

This branch is based on jseyfried's branch fixing the issue with bindgen (and clang?) (#88).


hanno-arm:
Ping @jethrogb and @jack-fortanix. 

The essence of @gbryant-arm's change is the introduction of Mbed TLS _as a submodule_, and while there's still some cleanup to do here on our side, it would be good to know if you're in general supportive of this change.

The rationale is that it will be easier to keep track of and update the version of Mbed TLS that the wrapper works with. We'd like this not only to ease support of upstream Mbed TLS, but also because we'd like to integrate the wrapper with an ongoing TLS 1.3 prototype implementation in Mbed TLS that we're working on.

What are your thoughts?

jethrogb:
Sorry for the late response. I haven't looked at the changes in this PR yet, but we actually did think about the desired architecture here a couple of months ago with some folks from Red Hat. We think something more akin to the way the [openssl-sys](https://crates.io/crates/openssl-sys) and [openssl-src](https://crates.io/crates/openssl-src) crates are organized is better. That way we can separate the bindings version from the MbedTLS version.

Taowyoo:
Close PR since long time no update

PR #100, opened 2020-04-15
Add unit test for certificate chain verification

jethrogb:


mbedtls/src/x509/certificate.rs
```diff
@@ -927,4 +927,30 @@ cYp0bH/RcPTC0Z+ZaqSWMtfxRrk63MJQF9EXpDCdvQRcTMD9D85DJrMKn8aumq0M
...
+        let mut c_leaf = Certificate::from_pem(C_LEAF.as_bytes()).unwrap();
+        let mut c_int1 = Certificate::from_pem(C_INT1.as_bytes()).unwrap();
+        let mut c_int2 = Certificate::from_pem(C_INT2.as_bytes()).unwrap();
+        let mut c_root = Certificate::from_pem(C_ROOT.as_bytes()).unwrap();
+
+        {
+            let mut chain = List::from(&mut c_leaf);
+            chain.push_back(&mut c_int1);
+            chain.push_back(&mut c_int2);
+            LinkedCertificate::verify((&mut chain).into(), &mut c_root, None).unwrap();
```
jack-fortanix, line 947: Suggest also a failing test even something simple like omitting one of the intermediates, just to make sure some kind of error is returned. As it is, a function implemented as `return true` would pass these tests.

jethrogb:
bors r=jack-fortanix

PR #101, opened 2020-04-16
Allow reusing an Md context object

jack-fortanix:
In cases where the `Md` is used once and thrown away, it is needless overhead. But the cost is small (`md_starts` is just resetting the chaining value and length field) and swamped by the cost of even 1 compression function. In cases where many hashes are computed (eg in LMS keygen) the overhead of creating a new struct adds up, and being able to reuse seems to help quite a bit.

### DIFF
--- a/mbedtls/src/hash/mod.rs
+++ b/mbedtls/src/hash/mod.rs
@@ -98,13 +98,14 @@ impl Md {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


Taowyoo:
Close this old PR because of change of default branch to `main`.
See more:
- https://github.com/fortanix/rust-mbedtls/issues/320

PR #103, opened 2020-05-07
Trusted Certificate Callback support

Pagten:
This PR adds support to rust-mbedtls for the `mbedtls_ssl_conf_ca_cb()` function. This function allows users to register the set of trusted certificates through a callback, instead of via a linked list as configured by `mbedtls_ssl_conf_ca_chain()`.


mbedtls/src/x509/certificate.rs
```diff
@@ -197,6 +197,17 @@ impl LinkedCertificate {
...
 
+    pub fn is_self_signed(&self) -> Result<bool> {
+        if self.inner.subject_raw.p == core::ptr::null_mut() || self.inner.subject_raw.len == 0 ||
+            self.inner.issuer_raw.p == core::ptr::null_mut() || self.inner.issuer_raw.len == 0 {
+            return Err(Error::X509BadInputData)
+        }
+
+        let subject = unsafe { core::slice::from_raw_parts(self.inner.subject_raw.p, self.inner.subject_raw.len) };
+        let issuer = unsafe { core::slice::from_raw_parts(self.inner.issuer_raw.p, self.inner.issuer_raw.len) };
+        Ok(subject == issuer)
```
jack-fortanix, line 208: This is not strictly correct because it will only find DNs which are bitwise identical. But the two DNs may actually be encoded in different ways (eg different ordering of DN components, different string types, and also IIRC the name comparisons are supposed to be be case-insensitive).

For many uses getting this wrong is not a huge problem (since in practice most CA certs are issued with identical DN bits). But for others it might cause a security problem since an attacker can construct a cert which some code will consider self-signed but this function will not consider self-signed, such as

```
peer_cert,extra_certs = parse_cert_list_sent_by_peer()
for cert in extra_certs:
   if cert.is_self_signed():
      return Error
   else:
      add_to_list_of_possible_intermediates(cert)
is_ok = validate(peer_cert)
```

If the peer sends a cert where issuer and subject DNs are identical but not bitwise identical, then this code would consider the cert not self signed, but other code (eg during certificate verification) might consider it to be self signed and treat it as a trust root.

Depending on the intended usage here I'm not sure if it is worth implementing all the required logic. Unfortunately mbedtls doesn't seem to expose a function for DN comparisons either. But at the least add a comment explaining why the logic here is incomplete and why it might cause problems re differing interpretations.

mbedtls-sys/Cargo.toml
```diff
@@ -50,3 +50,4 @@ aesni = []
 padlock = []
 legacy_protocols = []
 mpi_force_c_code = []
+trusted_cert_callback = []
```
jack-fortanix, line 53: It seems like enabling this feature in mbedtls adds very little additional code size. Could we just enable it always to avoid further proliferation of feature flags?

mbedtls/src/x509/certificate.rs
Pagten: You're right. The `is_self_signed()` function was a convenient and efficient way for me to perform the check that I need in application-level code, but - considering the concerns that you raised - it's not a function that should be part of the rust-mbedtls crate, at least not in its current form. I will remove this function and move the logic to the application.

mbedtls-sys/Cargo.toml
jethrogb: It's correct that this is a feature on mbedtls-sys. It could be turned by default in mbedtls though.

Pagten: Indeed, I've update the `Cargo.toml` in the mbedtls crate to enable the `trusted_cert_callback` feature in `mbedtls-sys-auto` by default.

mbedtls/tests/support/keys.rs
```diff
@@ -59,6 +59,83 @@ PVTB987Vm6hLu5/oHF+JVW05ZdXID1BZvukBYXnwY9OHvU7fin8N/eT/SBTd2HxO
 d1SSYr2U5pj0tNqaDQ==
 -----END CERTIFICATE-----\0";
 
+pub const PEM_SELF_SIGNED_CERT_INVALID_SIG: &'static [u8] = b"-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
Pagten: Will do

PR #109, opened 2020-05-26
Add accessors for RSA d, dp, dq, and qp

jack-fortanix:
We need this in order to expose these fields to Lua plugins (PROD-1660)


raoulstrackx:
Code looks good, but has a merge conflict

PR #110, opened 2020-06-03
Making Time struct accessible

raoulstrackx:
In order to properly verify certificates, their issue and expiration date needs to be accessible.


jethrogb:
Doesn't MbedTLS already take care of this?

PR #117, opened 2020-07-13
Raw Rsa decrypt Mode

nickedes:
This PR adds a new mode for Padding i.e. RawDecrypt.

### DIFF
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -38,7 +38,7 @@ pub use crate::ecp::EcGroup;
 // SHA-256("Fortanix")[:4]
 const CUSTOM_PK_TYPE: pk_type_t = 0x8b205408 as pk_type_t;
 
-const RAW_RSA_DECRYPT : i32 = 2147483647;
+const RAW_RSA_DECRYPT : i32 = 1040451858;
 
 define!(
     #[c_ty(pk_type_t)]
@@ -74,12 +74,12 @@ impl From<pk_type_t> for Type {
 
 pub enum RsaPadding {
     Pkcs1V15,
-    RawDecrypt,
     /// Use OAEP for encryption, or PSS for signing.
     Pkcs1V21 {
         /// The Mask Generating Function (MGF) to use.
         mgf: MdType,
     },
+    None,
 }
 
 pub enum Options {
@@ -295,7 +295,7 @@ impl Pk {
                 | (Type::RsassaPss, Options::Rsa { padding }) => {
                     let (padding, hash_id) = match padding {
                         RsaPadding::Pkcs1V15 => (RSA_PKCS_V15, 0),
-                        RsaPadding::RawDecrypt => {
+                        RsaPadding::None => {
                             let ctx = self.inner.pk_ctx as *mut rsa_context;
                             (*ctx).padding = RAW_RSA_DECRYPT; // denotes RawDecrypt padding being set
                             return;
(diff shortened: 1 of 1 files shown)


mbedtls/src/pk/mod.rs
```diff
@@ -38,6 +38,8 @@ pub use crate::ecp::EcGroup;
 // SHA-256("Fortanix")[:4]
 const CUSTOM_PK_TYPE: pk_type_t = 0x8b205408 as pk_type_t;
 
+const RAW_RSA_DECRYPT : i32 = 2147483647;
```
jethrogb, line 41: Please use a random value
```suggestion
const RAW_RSA_DECRYPT : i32 = 1040451858;
```

mbedtls/src/pk/mod.rs
```diff
@@ -72,6 +74,7 @@ impl From<pk_type_t> for Type {
 
 pub enum RsaPadding {
     Pkcs1V15,
+    RawDecrypt,
```
jethrogb, line 77:
```suggestion
    None,
```

mbedtls/src/pk/mod.rs
```diff
@@ -498,7 +506,21 @@ impl Pk {
         plain: &mut [u8],
         rng: &mut F,
     ) -> Result<usize> {
+        let ctx = self.inner.pk_ctx as *mut rsa_context;
```
jethrogb, line 509: You need to check this is an RSA key first

nickedes: This `self.pk_type() == Type::Rsa` check would be sufficient. Right?

mbedtls/src/pk/mod.rs
```diff
@@ -1064,6 +1088,135 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
...
+        let length_without_padding = pk.decrypt(&cipher, &mut decrypted_data1, &mut rng).unwrap();
+        // set raw decryption padding mode to perform raw decryption
+        pk.set_options(Options::Rsa {
+            padding: RsaPadding::None
+        });
+        let mut decrypted_data2 = [0u8; 2048 / 8];
+        let length_with_padding = pk.decrypt(&cipher, &mut decrypted_data2, &mut rng).unwrap();
+        // compare lengths of the decrypted texts
+        assert_ne!(length_without_padding, length_with_padding);
+        assert_ne!(&decrypted_data1[..], &decrypted_data2[..]);
```
jethrogb, line 1114: You should check assert_eq!(decrypted_data2.len(), cipher.len())

mbedtls/src/pk/mod.rs
```diff
@@ -1064,6 +1088,135 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
...
+            Type::Md5,
+            Type::Sha1,
+            Type::Sha224,
+            Type::Sha256,
+            Type::Sha384,
+            Type::Sha512,
+            Type::Ripemd,
+        ];
+
+        for digest in &digests {
```
jethrogb, line 1174: Testing every hash alg is a little excessive

mbedtls/src/pk/mod.rs
```diff
@@ -1064,6 +1088,135 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
...
+    }
+
+    #[test]
+    fn rsa_encrypt_with_no_padding() {
+        let mut pk = Pk::from_private_key(TEST_DER, None).unwrap();
+        let mut cipher = [0u8; 2048 / 8];
+        pk.set_options(Options::Rsa {
+            padding: RsaPadding::None
+        });
+        assert!(pk.encrypt(b"test", &mut cipher, &mut crate::test_support::rand::test_rng()).is_err());
```
jethrogb, line 1152: In general: use `result.unwrap_err()` instead of `assert!(result.is_err())`. In this case, please also check for the specific error code.

Also do this in the other tests.

mbedtls/src/pk/mod.rs
```diff
@@ -1064,6 +1088,135 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
...
+        });
+        let mut decrypted_data2 = [0u8; 2048 / 8];
+        let length_with_padding = pk.decrypt(&cipher, &mut decrypted_data2, &mut rng).unwrap();
+        // compare lengths of the decrypted texts
+        assert_ne!(length_without_padding, length_with_padding);
+        assert_ne!(&decrypted_data1[..], &decrypted_data2[..]);
+    }
+
+    #[test]
+    fn encrypt_raw_decrypt_with_pkcs1_v21() {
```
jethrogb, line 1118: This test is not necessary.

mbedtls/src/pk/mod.rs
jethrogb: Sorry for not being more clear the first time but I really meant that you can just use Sha256 only.

PR #119, opened 2020-07-15
Add `fn verify_result` in `ssl::Context`

mzohreva:
This would be helpful in getting more details about `Context::establish()` failures

### DIFF
--- a/mbedtls/src/ssl/context.rs
+++ b/mbedtls/src/ssl/context.rs
@@ -157,6 +157,13 @@ impl<'config> Context<'config> {
         }
     }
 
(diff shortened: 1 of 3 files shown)


Goirad:
Is it too late to start adding doc comments? A few lines about what is being returned would be nice I think. Besides that LGTM

mzohreva:
That is a general issue in this repo. I think it would be more appropriate to add comprehensive documentation in a separate PR.

Goirad:
Sounds good to me, I just hope we just aren't revisiting this topic a year from now

bors r+

PR #124, opened 2020-08-18
Fixes and improvements in Pk module

Pagten:
I noticed that the `decrypt()` function would take its return value from uninitialized memory when performing a raw RSA decrypt. Also, the length of the plaintext output buffer was not checked.

Edit: Some further changes were incorporated into this PR. The total list of changes is:
 * Fix returning uninitialized value from `Pk::decrypt()` when using RSA RAW mode
 * Remove unnecessary use of `MaybeUninit`
 * Add output buffer length check to `Pk::sign_deterministic()`

### DIFF
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -621,8 +621,11 @@ impl Pk {
         use crate::rng::RngCallback;
 
         if self.pk_type() == Type::Ecdsa || self.pk_type() == Type::Eckey {
-            // RFC 6979 signature scheme
+            if sig.len() < ECDSA_MAX_LEN {
+                return Err(Error::PkSigLenMismatch);
+            }
 
+            // RFC 6979 signature scheme
             let q = EcGroup::new(self.curve()?)?.order()?;
             let x = self.ec_private()?;
 
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -527,19 +527,18 @@ impl Pk {
             }
         }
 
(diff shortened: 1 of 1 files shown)


jethrogb:
> I noticed that the decrypt() function would take its return value from uninitialized memory when performing a raw RSA decrypt.

I don't see this?

jethrogb:
Upon closer inspection, I don't understand why this code is using MaybeUninit at all.

Pagten:
@jethrogb, are you suggesting to change the second half of `decrypt()` to:
```
        let mut ret = 0usize;
        unsafe {
            pk_decrypt(
                &mut self.inner,
                cipher.as_ptr(),
                cipher.len(),
                plain.as_mut_ptr(),
                &mut ret as *mut usize,
                plain.len(),
                Some(F::call),
                rng.data_ptr(),
            ).into_result()?
        };
        Ok(ret)
    }
```
?

jethrogb:
Yes and also encrypt, and can you see if there are any other obviously incorrect uses?

Pagten:
I'll do a check for other incorrect uses tomorrow. I think I'll create a separate PR for the MaybeUninit fixes, though, to keep a clean git history.

Pagten:
I decided to add the `MaybeUninit` changes to this PR after all, since stacking PRs across forks did not seem to work very well. I also noticed a missing output buffer length check in `sign_deterministic()`, so I added that check as well.

AdrianCX (approved):

Pagten:
bors r=AdrianCX

PR #127, opened 2020-10-13
Async I/O support + ALPN

mzohreva:


Pagten:
I went through the PR and it looks good to me, but I don't feel qualified to approve the async-related changes since I hardly have experience with async in rust. @jethrogb are you planning to review this?

jethrogb:
This will need significant changes after landing #128 which will come first

jethrogb:
Wrong target branch (or wrong rebase target)

mzohreva:
This should target v0.6 but 0.6 was master when I opened the PR I think. I'll close this PR for now.

mbedtls/src/ssl/async_session.rs
```diff
@@ -130,9 +130,81 @@ impl Drop for SessionWrapper<'_> {
...
+        out
+    }
+
+    fn adjust_buf<'a>(&self, buf: &'a [u8]) -> io::Result<&'a [u8]> {
+        match self.pending.as_ref() {
+            None => Ok(buf),
+            Some(pending) => {
+                if pending.len <= buf.len() {
+                    let buf = &buf[..pending.len];
+                    if Self::digest(buf) == pending.digest {
```
jethrogb, line 175: Do we really need to do this? It's expensive and unlikely to occur.

mzohreva: I agree that it is unlikely but if it happens it can cause a lot of trouble. Alternative would be to assume that the same bytes (with maybe more bytes appended) are written and just make sure we slice it to the previous length. And of course document the behavior, but we won't be able to throw an error if the conditions are not met. Do you think that's worth it?

raoulstrackx: Can't we do both? Do the expensive check when the crate is compiled in debug mode, and assume everything is correct in release mode?

mzohreva: I like that idea.

jethrogb:
Can this be rebased on 0.8?

mzohreva:
> Can this be rebased on 0.8?

I think it would not be straightforward to rebase this. I'll have to spend some time and do that separately.
But I think we can/should land this in 0.7 since it's being used as is.

raoulstrackx:
Is there any progress on this? It's blocking other work we need.

mbedtls/src/ssl/async_session.rs
zugzwang: Taking the liberty of resolving this (was implemented, see `debug_assertions`)

mbedtls/Cargo.toml
```diff
@@ -83,6 +84,16 @@ name = "server"
...
 
+[[test]]
+name = "alpn"
+path = "tests/alpn.rs"
+required-features = ["std"]
+
+[[test]]
+name = "async_session"
+path = "tests/async_session.rs"
+required-features = ["std", "threading", "tokio", "tokio/net", "tokio/io-util", "tokio/macros"]
```
zugzwang, line 95: Based on the below, I believe we should add `tokio/rt` to this list, and make sure the test actually runs in CI with `cargo test --features=std,threading,tokio,tokio/net,tokio/io-util,tokio/macros,tokio/rt`

Note 1: I did update to mbedtls-sys 2.28.0 to pick up https://github.com/fortanix/rust-mbedtls/pull/152 and being able to actually test. See the diff [here](https://github.com/fortanix/rust-mbedtls/pull/225) (a PR to this PR)

***
Note 2: For the following, added an `assert!(false)` to the very beginning of the `client_server_test` so it FAILS.

1. When I run `cargo test test::client_server_test` (no features), `cargo` says the test ran `ok` (I was expecting filtered out, or FAIL, so **this is either bad tokio usage in the test suite, or very misleading cargo behavior** (might be cargo, see Note above)  :x: 
2. When I run `cargo test test::client_server_test  --features=std,threading,tokio,tokio/net`, `cargo` says the test ran `ok` (so, **same as 1**) :x: 
3. When I run `cargo test test::client_server_test  --features=std,threading,tokio,tokio/net,tokio/io-util,tokio/macros` (the features specified here), I finally get a compilation error saying `The #[tokio::test] macro requires rt or rt-multi-thread.` :heavy_check_mark: 
4. When I run `cargo test test::client_server_test  --features=std,threading,tokio,tokio/net,tokio/io-util,tokio/macros,tokio/rt`, the test is finally ran, resulting in the expected FAILURE. :heavy_check_mark: 
5. When I add `tokio/rt` to the above list in line 95, the test is not run in my machine by any of the commands given in `ct.sh`, but they do parse it and report `ok` as if it had passed. :x:

Taowyoo:
bors try

zugzwang (commented):
The PR looks good to me, but I found misleading test behavior and provided a fix in https://github.com/fortanix/rust-mbedtls/pull/225 so that tokio tests actually run in 2.28.0 (see my comment).

Edit: I am not suggesting we update to 2.28.0 in this PR, only that we add the feature and the cargo command.

mbedtls/Cargo.toml
Taowyoo: Hi, I think the problem should because we have two tests both has name `test::client_server_test` , and the logic of using `cargo test TEST_NAME` is equals to run `cargo test` but also ask `cargo` to only run tests whose name includes `TEST_NAME`. And if you check the output of `cargo test test::client_server_test` there is no lines of `Running tests/async_session.rs ...`
The more accurate way is `cargo test --test async_session`
Then `cargo` will alert:
```
error: target `async_session` in package `mbedtls` requires the features: `std`, `threading`, `tokio`, `tokio/net`, `tokio/io-util`, `tokio/macros`, `tokio/rt`
Consider enabling them by passing, e.g., `--features="std threading tokio tokio/net tokio/io-util tokio/macros tokio/rt"
```

Taowyoo: Yes.
I updated related CI script in #225

zugzwang: Thanks.

zugzwang (approved):
Thanks for the PR.

After review and suggested changes addressed by @Taowyoo, this looks good to me.
I will give a tentative approval, subject to @mzohreva agreeing on https://github.com/fortanix/rust-mbedtls/pull/225.

I reviewed @Taowyoo's commits in https://github.com/fortanix/rust-mbedtls/pull/225. It includes:
- Updating to mbedtls-sys 2.28.0
- Make sure tests run on CI
- Fix CI and build

The plan (again, subject to @mzohreva agreeing to https://github.com/fortanix/rust-mbedtls/pull/225) is to either
- Approve definitely this PR, and decline it in favor of https://github.com/fortanix/rust-mbedtls/pull/224 which contains all the backported commits that succeeded this PR (everything reviewed). Merge 224.
- Incorporate the changes in https://github.com/fortanix/rust-mbedtls/pull/224 here (already reviewed) and merge 127.

Taowyoo:
> Thanks for the PR.
> 
> After review and suggested changes addressed by @Taowyoo, this looks good to me. I will give a tentative approval, subject to @mzohreva agreeing on #225.
> 
> I reviewed @Taowyoo's commits in #225. It includes:
> 
> * Updating to mbedtls-sys 2.28.0
> * Make sure tests run on CI
> * Fix CI and build
> 
> The plan (again, subject to @mzohreva agreeing to #225) is to either
> 
> * Approve definitely this PR, and decline it in favor of #224 which contains all the backported commits that succeeded this PR (everything reviewed). Merge 224.
> * Incorporate the changes in #224 here (already reviewed) and merge 127.
@zugzwang 
I think the second option is better, it will distinguish code change in this PR and #225 from #224.
So make it easy to review #224.

Issue #9, opened 2018-10-30
Check &self/&mut self


jethrogb:
There's a lot of *mut pointer usage in the MbedTLS C API but perhaps some of those usages should be *const, especially with regards to callbacks.

jethrogb:
> We could start assuming these functions are actually const and then avoid having to take &mut self by instead casting self.inner as *const _ as *mut _,

Yes that's what I was thinking

jack-fortanix:
Sounds good to me. Some of these will require some digging to figure out if it is actually safe or not (for instance it appears `ecdh_calc_secret` may actually modify the context). In https://github.com/ARMmbed/mbedtls/issues/1485#issuecomment-450443915 I suggested to upstream that they document the situation better.

Issue #8, opened 2018-10-30
Check Send/Sync


jethrogb:
We currently only implement Send on most structs when threading is enabled, but perhaps that restriction is unnecessary?

jethrogb:
Problem: SSL Context is only `Sync` if the RNG is `Sync`

Issue #4, opened 2018-10-30
Improve Config/Context API


jethrogb:
Currently the Config/Context types really only work well when you're able to put everything on a stack somewhere. Using Box/Arc requires lots of unsafe.

See also #3.

jethrogb:
The Rust OpenSSL bindings use the [foreign-types crate](https://crates.io/crates/foreign-types). We can consider using that instead of our own current wrapper macros.

Issue #3, opened 2018-10-30
Implement native-tls compatible APIs


jethrogb:
To make sure it's easy for people to swap in MbedTLS where they were using some other TLS library before, we need to make sure `native-tls` could be implemented in terms of MbedTLS.

jack-fortanix:
One problem will be, `native-tls` assumes that PKCS12 is used as the container for keys and certs. In fact it supports no other method of configuration (there is no way to say pass say a certificate chain + a PKCS8 private key). Mbedtls doesn't support PKCS12 at all.

Last year I worked on a project that used mbedtls that needed PKCS12, I hacked up something that happened to be able to generate valid PKCS12 files + parse them but it was fragile and (with mbedtls's quite primitive ASN1 library) very verbose, it was easily 1K lines of C.

If PKCS12 support proves required (as I think it would be for `native-tls`) it would be easier and safer to implement a PKCS12 library in Rust in a new crate, using `yasna` for ASN1 and `mbedtls` crate for the crypto ops.

jack-fortanix:
`rust-native-tls` requires that the stream types implement both `Sync` and `Send`. Already (with threading support enabled), `mbedtls::Context` is `Send`, but not `Sync`. This appears to be correct, in that the underlying `ssl_context` does not take a lock. I suppose the fix is, within `rust-native-tls`, add a mutex which serializes access to the `ssl_context` and then implement `Sync` for that wrapper type.

This assumes it is safe to invoke a shared `ssl_context` from multiple threads in a serialized way. Which it should be unless they are doing something weird, like using thread IDs for something, but as with the un-movable AES context sometimes mbedtls is kind of odd. So I'll review the code to make sure there are nothing like that.

jack-fortanix:
I'm clearly misunderstanding what `rust-native-tls` expects wrt `Sync`, `rust-openssl` defines the wrapper around `SSL` as `Sync` with no external lock, but OpenSSL's `SSL` is very much not safe to access concurrently from multiple threads.

jack-fortanix:
Things missing that native-tls expects

* Being able to distinguish server vs client endpoints. The getters in `Config` are commented out with "need bitfield support". This is needed to implement `tls_server_end_point` which returns a hash of the server certificate. This is used for channel binding.
* Explicit shutdown API. Right now, it happens as a result of the `Context` being `Drop`ed.

Things that are missing but I think I can work around:

[...]

PR #128, opened 2020-11-17
MbedTLS Reference counted instead of lifetimes

AdrianCX:
Moving from referene counting allows simpler move to native-tls / hyper.

Arc Changes:
- Each Config/Context/... will hold Arcs towards items it holds pointers to.
- This forces objects to live as long as needed, once no longer used they get destroyed by reference counting.

This allows passing the objects to multiple threads without worrying about lifetime.
I've also added notes why classes are Sync where used. Let me know if I missed any classes.

Usage example of an intermediate mbed-hyper integration is at: 
- https://github.com/fortanix/rust-mbedtls/tree/acruceru/wip-mbed-hyper-v2/mbedtls-hyper/examples/integrations

There I added a crate to wrap hyper - similar to native-tls. (that will be moved to native-tls layer soon)
That crate can be considered an integration test that I will raise a separate PR for.

Edit:

Changes after initial review:
-    Added forward_mbedtls_calloc / forward_mbedtls_free functions so we can pass certificates to and from mbedtls without allocator mismatches/corruptions.
-    Switched to MbedtlsList<Certificate> and Certificate. A MbedtlsBox is pending for this PR as well.
-    Fixed most comments.

Still pending:
-    Update define! macros
-    Add MbedtlsBox<Certificate>

Fixes https://github.com/fortanix/rust-mbedtls/issues/1
Partial progress on https://github.com/fortanix/rust-mbedtls/issues/3
Fixes https://github.com/fortanix/rust-mbedtls/issues/4
Fixes https://github.com/fortanix/rust-mbedtls/issues/8
Partially addresses https://github.com/fortanix/rust-mbedtls/issues/9

### DIFF
--- a/mbedtls-sys/Cargo.toml
+++ b/mbedtls-sys/Cargo.toml
@@ -12,6 +12,7 @@ This version generates the correct bindings at compile time using bindgen."""
 readme = "../README.md"
 repository = "https://github.com/fortanix/rust-mbedtls"
 documentation = "https://docs.rs/mbedtls-sys-auto/"
+links = "mbedtls"
 
 [lib]
 name = "mbedtls_sys"
--- a/mbedtls/src/lib.rs
+++ b/mbedtls/src/lib.rs
@@ -15,7 +15,7 @@ const ERROR: _MUST_USE_EITHER_STD_OR_CORE_IO_ = ();
 
 #[cfg(not(feature = "std"))]
 #[macro_use]
-extern crate alloc;
+extern crate alloc as rust_alloc;
 #[cfg(feature = "std")]
 extern crate core;
 #[cfg(not(feature = "std"))]
@@ -59,6 +59,7 @@ pub mod rng;
 pub mod self_test;
 pub mod ssl;
 pub mod x509;
+pub mod alloc;
 
 #[cfg(feature = "pkcs12")]
 pub mod pkcs12;
@@ -112,11 +113,13 @@ mod mbedtls {
 #[cfg(not(feature = "std"))]
 mod alloc_prelude {
     #![allow(unused)]
-    pub(crate) use alloc::borrow::ToOwned;
-    pub(crate) use alloc::boxed::Box;
-    pub(crate) use alloc::string::String;
-    pub(crate) use alloc::string::ToString;
-    pub(crate) use alloc::vec::Vec;
+    pub(crate) use rust_alloc::borrow::ToOwned;
+    pub(crate) use rust_alloc::boxed::Box;
+    pub(crate) use rust_alloc::sync::Arc;
+    pub(crate) use rust_alloc::string::String;
+    pub(crate) use rust_alloc::string::ToString;
+    pub(crate) use rust_alloc::vec::Vec;
+    pub(crate) use rust_alloc::borrow::Cow;
 }
 
 #[cfg(all(feature="time", any(feature="custom_gmtime_r", feature="custom_time")))]
@@ -163,3 +166,11 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut time_t) -> time_t {
     }
     timestamp
 }
+
+// Debug not available in SGX
+#[cfg(not(target_env = "sgx"))]
+pub unsafe fn set_global_debug_threshold(threshold: i32) {
+    mbedtls_sys::debug_set_threshold(threshold);
+}
+
+
--- a/mbedtls/src/pk/dhparam.rs
+++ b/mbedtls/src/pk/dhparam.rs
@@ -22,7 +22,7 @@ impl Dhm {
     /// Takes both DER and PEM forms of FFDH parameters in `DHParams` format.
     ///
     /// When calling on PEM-encoded data, `params` must be NULL-terminated
-    pub(crate) fn from_params(params: &[u8]) -> Result<Dhm> {
+    pub fn from_params(params: &[u8]) -> Result<Dhm> {
         let mut ret = Self::init();
         unsafe { dhm_parse_dhm(&mut ret.inner, params.as_ptr(), params.len()) }.into_result()?;
         Ok(ret)
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -35,6 +35,8 @@ pub use self::ec::{EcGroupId, ECDSA_MAX_LEN};
 #[doc(inline)]
 pub use crate::ecp::EcGroup;
 
+pub use dhparam::Dhm;
+
 // SHA-256("Fortanix")[:4]
 const CUSTOM_PK_TYPE: pk_type_t = 0x8b205408 as pk_type_t;
 
@@ -132,6 +134,7 @@ const CUSTOM_PK_INFO: pk_info_t = {
     }
 };
 
+// If this changes then certificate.rs unsafe code in public_key needs to also change.
 define!(
     #[c_ty(pk_context)]
     #[repr(C)]
@@ -141,6 +144,91 @@ define!(
     impl<'a> Into<ptr> {}
     impl<'a> UnsafeFrom<ptr> {}
 );
+
+// # Safety
+//
+// Thread safety analysis for Pk.
+//
+// A. Usage example of Pk.
+//
+// 1.1. Common use case is to to pass it as parameter to the SSL Config class.
+// 1.2. SSL Config class is then used by multiple Context classes (one for each connection)
+// 1.3. Context classes, handled by different threads will do calls towards Pk.
+//
+// Since this is a common use case for MbedTLS it should be thread safe if threading is enabled.
+//
+// B. Verifying thread safety.
+//
+// 1. Calls towards the specific Pk implementation are done via function pointers.
+// 
+// - Example call towards Pk:
+//    ../../../mbedtls-sys/vendor/library/ssl_srv.c:3707 - mbedtls_pk_decrypt( private_key, p, len, ...
+// - This calls a generic function pointer via:
+//    ../../../mbedtls-sys/vendor/crypto/library/pk.c:475 - return( ctx->pk_info->decrypt_func( ctx->pk_ctx, input, ilen,
+//
+// 2. Pk implementation types.
+//
+// - The function pointers are defined via function:
+//      ../../../mbedtls-sys/vendor/crypto/library/pk.c:115 - mbedtls_pk_info_from_type
+// - They are as follows: mbedtls_rsa_info / mbedtls_eckey_info / mbedtls_ecdsa_info
+// - These are defined in: 
+//       ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:196
+//
+// C. Checking types one by one.
+//
+// 1. RSA: mbedtls_rsa_info at ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:196
+// This uses internal locks in: ../../../mbedtls-sys/vendor/crypto/library/rsa.c:718
+//
+// 2. ECKEY: mbedtls_eckey_info at ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:418
+// This does not use internal locks but avoids interior mutability.
+//
+// Function checks one by one:
+// - Only const access to context: eckey_check_pair, eckey_get_bitlen, eckey_can_do, eckey_check_pair
+//
+// - Const acccess / copies context to a stack based variable
+//   eckey_verify_wrap, eckey_sign_wrap: ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:251
+//       creates a stack ecdsa variable and uses ctx to initialize it.
+//       ctx is passed as 'key', a const pointer to mbedtls_ecdsa_from_keypair( &ecdsa, ctx )
+//           ../../../mbedtls-sys/vendor/crypto/library/ecdsa.c:819
+//           int mbedtls_ecdsa_from_keypair( mbedtls_ecdsa_context *ctx, const mbedtls_ecp_keypair *key )
+//           key does not mutate.
+//
+// - Ignored due to not defined: eckey_verify_rs_wrap, eckey_sign_rs_wrap
+//   (Undefined - MBEDTLS_ECP_RESTARTABLE - ../../../mbedtls-sys/build/config.rs:173)
+//
+// - Only used when creating/freeing - which is safe by design - eckey_alloc_wrap / eckey_free_wrap
+//
+// 3. ECDSA: mbedtls_ecdsa_info at ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:729
+// This does not use internal locks but avoids interior mutability.
+//
+// - Const access / copies context to stack based variables:
+//   ecdsa_verify_wrap: ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:544
+//       This copies the public key on the stack - in buf[] and copies the group id and nbits.
+//       That is done via: mbedtls_pk_write_pubkey( &p, buf, &key ) where key.pk_ctx = ctx;
+//       And the key is a const parameter to mbedtls_pk_write_pubkey - ../../../mbedtls-sys/vendor/crypto/library/pkwrite.c:158
+//
+// - Const access with additional notes due to call stacks involved.
+//
+//   ecdsa_sign_wrap: ../../../mbedtls-sys/vendor/crypto/library/pk_wrap.c:657
+//       mbedtls_ecdsa_write_signature ../../../mbedtls-sys/vendor/crypto/library/ecdsa.c:688
+//           mbedtls_ecdsa_write_signature_restartable ../../../mbedtls-sys/vendor/crypto/library/ecdsa.c:640
+//               MBEDTLS_ECDSA_DETERMINISTIC is not defined.
+//               MBEDTLS_ECDSA_SIGN_ALT is not defined.
+//               Passes grp to: ecdsa_sign_restartable: ../../../mbedtls-sys/vendor/crypto/library/ecdsa.c:253
+//                    Const access to group - reads parameters, passed as const to mbedtls_ecp_gen_privkey,
+//                    mbedtls_ecp_mul_restartable: ../../../mbedtls-sys/vendor/crypto/library/ecp.c:2351
+//                        MBEDTLS_ECP_INTERNAL_ALT is not defined. (otherwise it might not be safe depending on ecp_init/ecp_free) ../../../mbedtls-sys/build/config.rs:131
+//                        Passes as const to: mbedtls_ecp_check_privkey / mbedtls_ecp_check_pubkey / mbedtls_ecp_get_type( grp
+//        
+// - Ignored due to not defined: ecdsa_verify_rs_wrap, ecdsa_sign_rs_wrap, ecdsa_rs_alloc, ecdsa_rs_free
+//   (Undefined - MBEDTLS_ECP_RESTARTABLE - ../../../mbedtls-sys/build/config.rs:173)
+//
+// - Only const access to context: eckey_check_pair
+//
+// - Only used when creating/freeing - which is safe by design: ecdsa_alloc_wrap, ecdsa_free_wrap
+//
+#[cfg(feature = "threading")]
+unsafe impl Sync for Pk {}
 
 impl Pk {
     /// Takes both DER and PEM forms of PKCS#1 or PKCS#8 encoded keys.
@@ -778,7 +866,7 @@ impl Pk {
         sig: &mut [u8],
         rng: &mut F,
     ) -> Result<usize> {
-        use crate::rng::RngCallback;
+        use crate::rng::RngCallbackMut;
 
         if self.pk_type() == Type::Ecdsa || self.pk_type() == Type::Eckey {
             if sig.len() < ECDSA_MAX_LEN {
@@ -803,8 +891,8 @@ impl Pk {
                     hash.len(),
                     sig.as_mut_ptr(),
                     &mut ret,
-                    Some(Rfc6979Rng::call),
-                    rng.data_ptr(),
+                    Some(Rfc6979Rng::call_mut),
+                    rng.data_ptr_mut(),
                 ).into_result()?;
             };
             Ok(ret)
--- a/mbedtls/src/pk/rfc6979.rs
+++ b/mbedtls/src/pk/rfc6979.rs
@@ -12,7 +12,7 @@ use crate::alloc_prelude::*;
 use mbedtls_sys::types::raw_types::{c_int, c_uchar, c_void};
 use mbedtls_sys::types::size_t;
 
-use crate::rng::{HmacDrbg, Random, RngCallback};
+use crate::rng::{HmacDrbg, Random, RngCallbackMut};
 
 use crate::error::Result;
 use crate::bignum::Mpi;
@@ -67,7 +67,7 @@ fn generate_rfc6979_nonce(md: &MdInfo, x: &Mpi, q: &Mpi, digest_bytes: &[u8]) ->
 pub(crate) struct Rfc6979Rng {
     pub k: Vec<u8>,
     pub k_read: usize,
-    pub rng: HmacDrbg<'static>,
+    pub rng: HmacDrbg,
 }
 
 /// An RNG which first outputs the k for RFC 6797 followed by random data
@@ -110,8 +110,8 @@ impl Rfc6979Rng {
     }
 }
 
-impl RngCallback for Rfc6979Rng {
-    unsafe extern "C" fn call(
+impl RngCallbackMut for Rfc6979Rng {
+    unsafe extern "C" fn call_mut(
         user_data: *mut c_void,
         data_ptr: *mut c_uchar,
         len: size_t,
@@ -126,7 +126,7 @@ impl RngCallback for Rfc6979Rng {
         }
     }
 
-    fn data_ptr(&mut self) -> *mut c_void {
+    fn data_ptr_mut(&mut self) -> *mut c_void {
         self as *const _ as *mut _
     }
 }
--- a/mbedtls/src/rng/ctr_drbg.rs
+++ b/mbedtls/src/rng/ctr_drbg.rs
@@ -6,126 +6,67 @@
  * option. This file may not be copied, modified, or distributed except
  * according to those terms. */
 
+#[cfg(feature = "std")]
+use std::sync::Arc;
+
+pub use mbedtls_sys::CTR_DRBG_RESEED_INTERVAL as RESEED_INTERVAL;
+use mbedtls_sys::*;
 use mbedtls_sys::types::raw_types::{c_int, c_uchar, c_void};
 use mbedtls_sys::types::size_t;
-pub use mbedtls_sys::CTR_DRBG_RESEED_INTERVAL as RESEED_INTERVAL;
-use mbedtls_sys::{
-    ctr_drbg_random, ctr_drbg_reseed, ctr_drbg_seed, ctr_drbg_set_prediction_resistance,
-    ctr_drbg_update, CTR_DRBG_PR_OFF, CTR_DRBG_PR_ON,
-};
 
-use super::{EntropyCallback, RngCallback};
-use crate::error::{IntoResult, Result};
-
-// ==== BEGIN IMMOVABLE TYPE KLUDGE ====
-// `ctr_drbg_context` inlines an `aes_context`, which is immovable. See
-// https://github.com/ARMmbed/mbedtls/issues/2147. We work around this
-// by always boxing up the context, which requires this module to depend on
-// std/alloc.
-//
-// If `ctr_drbg_context` were moveable, this entire section could be replaced
-// by basically:
-// ```
-// define!(
-//     #[c_ty(ctr_drbg_context)]
-//     struct CtrDrbg<'entropy>;
-//     fn init() {
-//         ctr_drbg_init
-//     }
-//     fn drop() {
-//         ctr_drbg_free
-//     }
-// );
-// ```
-
-use self::private::CtrDrbgInner;
 #[cfg(not(feature = "std"))]
 use crate::alloc_prelude::*;
-use core::ops::{Deref, DerefMut};
+use crate::error::{IntoResult, Result};
+use crate::rng::{EntropyCallback, RngCallback, RngCallbackMut};
 
-mod private {
-    use core::marker::PhantomData;
-    use mbedtls_sys::{ctr_drbg_context, ctr_drbg_free, ctr_drbg_init};
+define!(
+    // `ctr_drbg_context` inlines an `aes_context`, which is immovable. See
+    // https://github.com/ARMmbed/mbedtls/issues/2147. We work around this
+    // by always boxing up the context, which requires this module to depend on
+    // std/alloc.
+    //
+    // If `ctr_drbg_context` were moveable we could use c_ty instead of c_box_ty.
+    //
+    #[c_box_ty(ctr_drbg_context)]
+    #[repr(C)]
+    struct CtrDrbg {
+        entropy: Arc<dyn EntropyCallback + 'static>,
+    };
+    const drop: fn(&mut Self) = ctr_drbg_free;
+    impl<'a> Into<ptr> {}
+);
 
-    pub struct CtrDrbgInner<'entropy> {
-        pub(super) inner: ctr_drbg_context,
-        r: PhantomData<&'entropy ()>,
-    }
-
-    impl<'entropy> CtrDrbgInner<'entropy> {
-        pub(super) fn init() -> Self {
-            let mut inner = ::core::mem::MaybeUninit::uninit();
-            let inner = unsafe {
-                ctr_drbg_init(inner.as_mut_ptr());
-                inner.assume_init()
-            };
-            CtrDrbgInner {
-                inner,
-                r: PhantomData,
-            }
-        }
-    }
-
-    impl<'entropy> Drop for CtrDrbgInner<'entropy> {
-        fn drop(&mut self) {
-            unsafe { ctr_drbg_free(&mut self.inner) };
-        }
-    }
-}
-
-pub struct CtrDrbg<'entropy> {
-    boxed: Box<CtrDrbgInner<'entropy>>,
-}
-
-impl<'entropy> CtrDrbg<'entropy> {
-    fn init() -> Self {
-        CtrDrbg {
-            boxed: Box::new(CtrDrbgInner::init()),
-        }
-    }
-}
-
-#[doc(hidden)]
-impl<'entropy> Deref for CtrDrbg<'entropy> {
-    type Target = CtrDrbgInner<'entropy>;
-    fn deref(&self) -> &Self::Target {
-        &self.boxed
-    }
-}
-
-#[doc(hidden)]
-impl<'entropy> DerefMut for CtrDrbg<'entropy> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.boxed
-    }
-}
-
-// ==== END IMMOVABLE TYPE KLUDGE ====
-
+//
+// Class has interior mutability via function called 'call'.
+// That function has an internal mutex to guarantee thread safety.
+//
+// The other potential conflict is a mutable reference changing class.
+// That is avoided by having any users of the callback hold an 'Arc' to this class.
+// Rust will then ensure that a mutable reference cannot be aquired if more then 1 Arc exists to the same class.
+//
 #[cfg(feature = "threading")]
-unsafe impl<'entropy> Sync for CtrDrbg<'entropy> {}
+unsafe impl Sync for CtrDrbg {}
 
-impl<'entropy> CtrDrbg<'entropy> {
-    pub fn new<F: EntropyCallback>(
-        source: &'entropy mut F,
-        additional_entropy: Option<&[u8]>,
-    ) -> Result<CtrDrbg<'entropy>> {
-        let mut ret = Self::init();
+#[allow(dead_code)]
+impl CtrDrbg {
+
+    pub fn new<T: EntropyCallback + 'static>(entropy: Arc<T>, additional_entropy: Option<&[u8]>) -> Result<Self> {
+        let mut inner = Box::new(ctr_drbg_context::default());
+
         unsafe {
+            ctr_drbg_init(&mut *inner);
             ctr_drbg_seed(
-                &mut ret.inner,
-                Some(F::call),
-                source.data_ptr(),
-                additional_entropy
-                    .map(<[_]>::as_ptr)
-                    .unwrap_or(::core::ptr::null()),
+                &mut *inner,
+                Some(T::call),
+                entropy.data_ptr(),
+                additional_entropy.map(<[_]>::as_ptr).unwrap_or(::core::ptr::null()),
                 additional_entropy.map(<[_]>::len).unwrap_or(0)
-            )
-            .into_result()?
-        };
-        Ok(ret)
-    }
+            ).into_result()?;
+        }
 
+        Ok(CtrDrbg { inner, entropy })
+    }
+    
     pub fn prediction_resistance(&self) -> bool {
         if self.inner.prediction_resistance == CTR_DRBG_PR_OFF {
             false
@@ -137,7 +78,7 @@ impl<'entropy> CtrDrbg<'entropy> {
     pub fn set_prediction_resistance(&mut self, pr: bool) {
         unsafe {
             ctr_drbg_set_prediction_resistance(
-                &mut self.inner,
+                &mut *self.inner,
                 if pr { CTR_DRBG_PR_ON } else { CTR_DRBG_PR_OFF },
             )
         }
@@ -151,7 +92,7 @@ impl<'entropy> CtrDrbg<'entropy> {
     pub fn reseed(&mut self, additional_entropy: Option<&[u8]>) -> Result<()> {
         unsafe {
             ctr_drbg_reseed(
-                &mut self.inner,
+                &mut *self.inner,
                 additional_entropy
                     .map(<[_]>::as_ptr)
                     .unwrap_or(::core::ptr::null()),
@@ -163,7 +104,7 @@ impl<'entropy> CtrDrbg<'entropy> {
     }
 
     pub fn update(&mut self, entropy: &[u8]) {
-        unsafe { ctr_drbg_update(&mut self.inner, entropy.as_ptr(), entropy.len()) };
+        unsafe { ctr_drbg_update(&mut *self.inner, entropy.as_ptr(), entropy.len()) };
     }
 
     // TODO:
@@ -174,13 +115,26 @@ impl<'entropy> CtrDrbg<'entropy> {
     //
 }
 
-impl<'entropy> RngCallback for CtrDrbg<'entropy> {
+impl RngCallbackMut for CtrDrbg {
     #[inline(always)]
-    unsafe extern "C" fn call(user_data: *mut c_void, data: *mut c_uchar, len: size_t) -> c_int {
+    unsafe extern "C" fn call_mut(user_data: *mut c_void, data: *mut c_uchar, len: size_t) -> c_int where Self: Sized {
+        // Mutex used in ctr_drbg_random at: ../../../mbedtls-sys/vendor/crypto/library/ctr_drbg.c:546
         ctr_drbg_random(user_data, data, len)
     }
 
-    fn data_ptr(&mut self) -> *mut c_void {
-        &mut self.inner as *mut _ as *mut _
+    fn data_ptr_mut(&mut self) -> *mut c_void {
+        self.handle_mut() as *const _ as *mut _
+    }
+}
+
+impl RngCallback for CtrDrbg {
+    #[inline(always)]
+    unsafe extern "C" fn call(user_data: *mut c_void, data: *mut c_uchar, len: size_t) -> c_int where Self: Sized {
+        // Mutex used in ctr_drbg_random at: ../../../mbedtls-sys/vendor/crypto/library/ctr_drbg.c:546
+        ctr_drbg_random(user_data, data, len)
+    }
+    
+    fn data_ptr(&self) -> *mut c_void {
+        self.handle() as *const _ as *mut _
     }
 }
--- a/mbedtls/src/rng/hmac_drbg.rs
+++ b/mbedtls/src/rng/hmac_drbg.rs
@@ -6,44 +6,53 @@
  * option. This file may not be copied, modified, or distributed except
  * according to those terms. */
 
+
+#[cfg(feature = "std")]
+use std::sync::Arc;
+
+pub use mbedtls_sys::HMAC_DRBG_RESEED_INTERVAL as RESEED_INTERVAL;
+use mbedtls_sys::*;
 use mbedtls_sys::types::raw_types::{c_int, c_uchar, c_void};
 use mbedtls_sys::types::size_t;
-pub use mbedtls_sys::HMAC_DRBG_RESEED_INTERVAL as RESEED_INTERVAL;
-use mbedtls_sys::{
-    hmac_drbg_random, hmac_drbg_reseed, hmac_drbg_seed, hmac_drbg_seed_buf,
-    hmac_drbg_set_prediction_resistance, hmac_drbg_update, HMAC_DRBG_PR_OFF, HMAC_DRBG_PR_ON,
-};
 
-use super::{EntropyCallback, RngCallback};
+#[cfg(not(feature = "std"))]
+use crate::alloc_prelude::*;
 use crate::error::{IntoResult, Result};
 use crate::hash::MdInfo;
+use crate::rng::{EntropyCallback, RngCallback, RngCallbackMut};
 
 define!(
     #[c_ty(hmac_drbg_context)]
-    struct HmacDrbg<'entropy>;
-    const init: fn() -> Self = hmac_drbg_init;
+    struct HmacDrbg {
+        entropy: Option<Arc<dyn EntropyCallback + 'static>>,
+    };
     const drop: fn(&mut Self) = hmac_drbg_free;
+    impl<'a> Into<ptr> {}
 );
 
 #[cfg(feature = "threading")]
-unsafe impl<'entropy> Sync for HmacDrbg<'entropy> {}
+unsafe impl Sync for HmacDrbg {}
 
-impl<'entropy> HmacDrbg<'entropy> {
-    pub fn new<F: EntropyCallback>(
+impl HmacDrbg {
+    pub fn new<T: EntropyCallback + 'static>(
         md_info: MdInfo,
-        source: &'entropy mut F,
+        entropy: Arc<T>,
         additional_entropy: Option<&[u8]>,
-    ) -> Result<HmacDrbg<'entropy>> {
-        let mut ret = Self::init();
+    ) -> Result<HmacDrbg> {
+
+        let mut ret = HmacDrbg {
+            inner: hmac_drbg_context::default(),
+            entropy: Some(entropy),
+        };
+        
         unsafe {
+            hmac_drbg_init(&mut ret.inner);
             hmac_drbg_seed(
                 &mut ret.inner,
                 md_info.into(),
-                Some(F::call),
-                source.data_ptr(),
-                additional_entropy
-                    .map(<[_]>::as_ptr)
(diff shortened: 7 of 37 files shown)


mbedtls/src/rng/hmac_drbg.rs
```diff
@@ -9,60 +9,79 @@
...
-    const init: fn() -> Self = hmac_drbg_init;
-    const drop: fn(&mut Self) = hmac_drbg_free;
-);
+#[allow(dead_code)]
+pub struct HmacDrbg {
+    // Moving data causes dangling pointers: https://github.com/ARMmbed/mbedtls/issues/2147
+    // Storing data in heap and forcing rust move to only move the pointer (box) referencing it.
+    // The move will be faster. Access to data will be slower due to additional indirection.
+    inner: Box<hmac_drbg_context>,
+    entropy: Option<Arc<dyn EntropyCallback + 'static>>,
```
jethrogb, line 25: I think we should allow the user to choose whether they use Arc or Box, for example like so:
```rust
use std::sync::Arc;

// somewhere in mbedtls public interface
pub trait IntoPinned: private::IntoPinnedPrivate {}
impl<T> IntoPinned for Box<T> {}
impl<T> IntoPinned for Arc<T> {}

// in mbedtls::private
mod private {
    use std::ops::Deref;
    use std::sync::Arc;

    enum PinnedStorage<T> {
        Box(Box<T>),
        Arc(Arc<T>),
    }

    impl<T> Deref for PinnedStorage<T> {
        type Target = T;
        
        fn deref(&self) -> &T {
            match self {
                PinnedStorage::Box(v) => &**v,
                PinnedStorage::Arc(v) => &**v,
            }
        }
    }

    pub trait IntoPinnedPrivate {
        type Contained;

        fn into_pinned(self) -> PinnedStorage<Self::Contained>;
    }

    impl<T> IntoPinnedPrivate for Box<T> {
        type Contained = T;

        fn into_pinned(self) -> PinnedStorage<T> {
            PinnedStorage::Box(self)
        }
    }

    impl<T> IntoPinnedPrivate for Arc<T> {
        type Contained = T;

        fn into_pinned(self) -> PinnedStorage<T> {
            PinnedStorage::Arc(self)
        }
    }
}
```

jethrogb (commented):
Not done with the review, but let's start here

mbedtls-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mbedtls-sys-auto"
-version = "2.18.5"
+version = "2.19.0"
```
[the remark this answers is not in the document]
AdrianCX: thanks, fixed.

mbedtls-sys/vendor/include/mbedtls/ssl.h
```diff
@@ -2404,8 +2404,8 @@ void mbedtls_ssl_conf_cert_profile( mbedtls_ssl_config *conf,
  * \param ca_crl   trusted CA CRLs
  */
 void mbedtls_ssl_conf_ca_chain( mbedtls_ssl_config *conf,
-                               mbedtls_x509_crt *ca_chain,
-                               mbedtls_x509_crl *ca_crl );
+                               const mbedtls_x509_crt *ca_chain,
+                               const mbedtls_x509_crl *ca_crl );
```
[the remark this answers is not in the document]
AdrianCX: thanks, removed these and added some into traits that return mutable pointers.

mbedtls-sys/build/config.rs
```diff
@@ -200,7 +200,7 @@ pub const DEFAULT_DEFINES: &'static [CDefine] = &[
     ("MBEDTLS_PSA_CRYPTO_SPM",                            Undefined),
     ("MBEDTLS_PSA_INJECT_ENTROPY",                        Undefined),
     ("MBEDTLS_RSA_NO_CRT",                                Undefined),
-    ("MBEDTLS_SELF_TEST",                                 Defined),
+    ("MBEDTLS_SELF_TEST",                                 Undefined),
```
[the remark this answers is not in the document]
AdrianCX: Added back and added a warning comment in self_test.rs. (we're exporting the mbedtls functions directly, I can add wrappers and a much more visible warning if needed)

mbedtls/src/wrapper_macros.rs
```diff
@@ -18,42 +18,104 @@ macro_rules! callback {
...
+            unsafe extern "C" fn call(user_data: *mut ::mbedtls_sys::types::raw_types::c_void, $($arg:$ty),*) -> $ret where Self: Sized {
+                (&mut*(user_data as *mut F))($($arg),*)
+            }
+
+            fn data_ptr(&self) -> *mut ::mbedtls_sys::types::raw_types::c_void {
+                self as *const F as *mut _
             }
         }
+    };
+    { $n:ident, $m:ident$( : $sync:ident )*($($arg:ident: $ty:ty),*) -> $ret:ty } => {
```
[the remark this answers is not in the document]
AdrianCX: Updated as suggested, let me know if I missed anything.

mbedtls/src/wrapper_macros.rs
```diff
@@ -225,45 +287,3 @@ macro_rules! define_struct {
     { << $name:ident $(lifetime $l:tt)* inner $inner:ident >> } => {};
     { lifetime $l:tt } => {};
 }
-
-macro_rules! setter {
```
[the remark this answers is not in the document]
AdrianCX: They do not provide much value, the get/set functions are one line with or without macro.

Function with macro:
```
    getter!(pk_type() -> Type = fn pk_get_type);
```
Function without macro:
```
    pub fn pk_type(&self) -> Type {
        unsafe { pk_get_type(&self.inner).into() }
    }
```

I can put back the macro if you want but for me it makes it harder to reason about what function does compared to the actual code.

mbedtls/src/pk/dhparam.rs
```diff
@@ -19,6 +19,11 @@ define!(
 );
 
 impl Dhm {
+    // For use when handling Arc<Dhm>
+    pub fn inner_ptr_const(&self) -> *const dhm_context {
```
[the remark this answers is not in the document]
AdrianCX: Removed all inner_ptr functions and switched to into traits.

mbedtls/src/pk/mod.rs
```diff
@@ -142,7 +143,99 @@ define!(
...
+unsafe impl Sync for Pk {}
+
 impl Pk {
+    pub fn inner_ptr(&mut self) -> *mut pk_context {
+        &mut self.inner
+    }
+    
+    pub fn inner_ptr_const(&self) -> *const pk_context {
+        &self.inner
+    }
```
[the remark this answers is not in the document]
AdrianCX: Removed all inner_ptr functions and switched to into traits.

mbedtls/src/pk/rfc6979.rs
```diff
@@ -87,7 +87,7 @@ impl Rfc6979Rng {
         let k = generate_rfc6979_nonce(&md, x, q, digest_bytes)?;
 
         Ok(Rfc6979Rng {
-            k: k,
+             k: k,
```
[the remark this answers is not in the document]
AdrianCX: fixed

mbedtls/src/rng/hmac_drbg.rs
```diff
@@ -9,60 +9,79 @@
...
-    struct HmacDrbg<'entropy>;
-    const init: fn() -> Self = hmac_drbg_init;
-    const drop: fn(&mut Self) = hmac_drbg_free;
-);
+#[allow(dead_code)]
+pub struct HmacDrbg {
+    // Moving data causes dangling pointers: https://github.com/ARMmbed/mbedtls/issues/2147
+    // Storing data in heap and forcing rust move to only move the pointer (box) referencing it.
+    // The move will be faster. Access to data will be slower due to additional indirection.
+    inner: Box<hmac_drbg_context>,
```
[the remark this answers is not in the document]
AdrianCX: double checked and you are right for this. Removed the box.

mbedtls/src/wrapper_macros.rs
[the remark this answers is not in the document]
jethrogb: The macro is used to avoid having to reason about all the unsafe blocks that would otherwise be introduced

mbedtls/src/lib.rs
```diff
@@ -7,6 +7,7 @@
  * according to those terms. */
 
 #![deny(warnings)]
+#![allow(dead_code)]
```
jethrogb, line 10: Why?

mbedtls/src/pk/dhparam.rs
```diff
@@ -18,6 +18,12 @@ define!(
     impl<'a> Into<ptr> {}
 );
 
+impl Into<*mut mbedtls_sys::dhm_context> for &Dhm {
```
jethrogb, line 21: If you think this is useful-it's a bit odd conversion?-please implement this as part of the define macro.

mbedtls/src/pk/mod.rs
```diff
@@ -22,7 +22,7 @@ use crate::hash::Type as MdType;
 use byteorder::{BigEndian, ByteOrder};
 
 pub(crate) mod dhparam;
-mod ec;
+pub mod ec;
```
jethrogb, line 25: I don't think we want to make this public.

mbedtls/src/rng/hmac_drbg.rs
```diff
@@ -9,51 +9,70 @@
...
-    #[c_ty(hmac_drbg_context)]
-    struct HmacDrbg<'entropy>;
-    const init: fn() -> Self = hmac_drbg_init;
-    const drop: fn(&mut Self) = hmac_drbg_free;
-);
+#[allow(dead_code)]
+pub struct HmacDrbg {
+    // Moving data causes dangling pointers: https://github.com/ARMmbed/mbedtls/issues/2147
+    // Storing data in heap and forcing rust move to only move the pointer (box) referencing it.
+    // The move will be faster. Access to data will be slower due to additional indirection.
```
jethrogb, line 23: This comment seems stale?

mbedtls/src/rng/hmac_drbg.rs
```diff
@@ -9,51 +9,70 @@
...
+    inner: hmac_drbg_context,
```
jethrogb, line 24: Please use the define macro

mbedtls/src/rng/ctr_drbg.rs
```diff
@@ -6,126 +6,76 @@
  * option. This file may not be copied, modified, or distributed except
  * according to those terms. */
 
+use mbedtls_sys::*;
+use std::sync::Arc;
 use mbedtls_sys::types::raw_types::{c_int, c_uchar, c_void};
 use mbedtls_sys::types::size_t;
-pub use mbedtls_sys::CTR_DRBG_RESEED_INTERVAL as RESEED_INTERVAL;
```
jethrogb, line 11: Why are you removing the reexport?

mbedtls/src/rng/rdrand.rs
```diff
@@ -64,30 +64,39 @@ fn write_rng_to_slice(outbuf: &mut [u8], rng: fn() -> Option<usize>) -> c_int {
...
         write_rng_to_slice(&mut outbuf, rdrand)
     }
 
-    fn data_ptr(&mut self) -> *mut c_void {
+    fn data_ptr_mut(&mut self) -> *mut c_void {
         ::core::ptr::null_mut()
     }
 }
+
+impl RngCallback for Rdrand {
```
jethrogb, line 98: I'm confused, where is `fn call` in this impl?

mbedtls/src/self_test.rs
```diff
@@ -58,6 +58,11 @@ pub unsafe fn disable() {
     log_f = None;
 }
 
+//
+// NOTE: The following functions if used in a multithreaded or async environment will fail.
+//       Self test functions rely on global variables to track operations and anything non-self-test related will stomp over variables.
+//       While using these, make sure no other code uses mbedtls. Multiple self test operations done asynchronously may also return failures.
+//
```
jethrogb, line 65: This note only makes sense as a doc comment. Please also make change the comments to `fn enable/disable` to doc comments. Follow this template:
```
/// # Safety
///
/// text here
```

mbedtls/src/ssl/ciphersuites.rs
```diff
@@ -9,6 +9,7 @@
 use mbedtls_sys::types::raw_types::c_int;
 use mbedtls_sys::*;
 
+// Always use into() to convert to i32, do not use 'as i32'
```
jethrogb, line 12: Is this intended as a doc comment? And why is not ok to type-cast? And conversion should happen to `c_int`, not `i32`.

mbedtls/src/x509/mod.rs
```diff
@@ -10,7 +10,7 @@
 use crate::alloc_prelude::*;
 
 pub mod certificate;
-mod crl;
```
jethrogb, line 13: Why are you making this change?

jethrogb (commented):
Add the following text in the issue description (if appropriate):
Fixes #1
Fixes #3
Fixes #4
Fixes #8
Fixes #9

mbedtls/src/x509/certificate.rs
```diff
@@ -61,87 +85,142 @@ impl BERDecodable for Extension {
...
+            }
+        }
     }
 }
 
-#[repr(C)]
-pub struct LinkedCertificate {
-    inner: x509_crt,
+pub struct LinkedCertificate<'a> {
+    pub inner: &'a x509_crt
```
jethrogb, line 223: Why are you changing this type?

AdrianCX: I do not want stack copies of the certificate as that makes possibly unsafe code.

With reference and lifetime I can make sure that LinkedCertificate only exists as long as parent certificate exists.
And since parent certificate makes sure nothing goes out of scope / gets delete - this ensures we are memory safe.

mbedtls/src/x509/mod.rs
AdrianCX: mistake, I'll revert

mbedtls/src/x509/certificate.rs
jethrogb: There are not stack copies. We only ever create references to LinkedCertificate.

mbedtls/src/rng/hmac_drbg.rs
AdrianCX: I need an additional field so I cannot use the macro.
```
    entropy: Option<Arc<dyn EntropyCallback + 'static>>,
```

jethrogb: There are lots of fields added throughout now with your new scheme, so please adjust the macro to make this possible.

mbedtls/src/x509/certificate.rs
AdrianCX: This is how I create it now:
```
    pub fn linked_cert(&self) -> LinkedCertificate<'_> {
        LinkedCertificate { inner: &*self.inner }
    }
```

This is how it was created in the past:
```
impl Deref for Certificate {
    type Target = LinkedCertificate;
    fn deref(&self) -> &LinkedCertificate {
        unsafe { UnsafeFrom::from(&self.inner as *const _).unwrap() }
    }
}
```

So you are correct it only ever created references in the past.
But I'm reluctant to use the unsafe / UnsafeFrom. Also the Certificate may have a different structure.

jethrogb: This was working correctly and reviewed. Please justify the change.

AdrianCX: It is simple to understand at first glance what this does:
```
    pub fn linked_cert(&self) -> LinkedCertificate<'_> {
        LinkedCertificate { inner: &*self.inner }
    }
```

Compared to old approach I linked above.

Let me know if I should revert.

jethrogb: This is not a justification. If the change is not needed, then please revert yes.

mbedtls/src/lib.rs
AdrianCX: Some variables are only set and never read. 

I moved from this define to those variables -> see mbedtls/src/ssl/config.rs:122

```
    #[allow(dead_code)]
    dhm: Option<Arc<Dhm>>,
```

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+
+#[cfg(feature = "threading")]
+unsafe impl<> Sync for MbedtlsBox<Certificate> {}
+
+impl<> MbedtlsBox<Certificate> {
+    pub fn new() -> Result<Self> {
+        unsafe {
+            let inner = forward_mbedtls_calloc(1, core::mem::size_of::<x509_crt>()) as *mut x509_crt;
+            if inner == core::ptr::null_mut() {
+                return Err(Error::SslAllocFailed);
```
jethrogb, line 39: X509AllocFailed

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            x509_crt_init(inner);
+            Ok(Self { inner: NonNull::new_unchecked(inner as *mut Certificate) })
```
jethrogb, line 43: 1. return the right error code
2. check alignment before constructing NonNull
3. don't use `Self` to construct a struct
4. don't check and then use the unchecked constructor

All combined:
```rust
let inner = forward_mbedtls_calloc(1, core::mem::size_of::<x509_crt>()) as *mut x509_crt;
assert_eq!(inner.align_offset(core::mem::align_of::<x509_crt>()), 0);
let inner = NonNull::new(inner).ok_or(Error::SslAllocFailed)?;
x509_crt_init(inner.as_ptr());
Ok(MbedtlsBox { inner: inner.cast() })
```

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+impl Into<*mut x509_crt> for &MbedtlsBox<Certificate> {
+    fn into(self) -> *mut x509_crt {
+        self.inner.as_ptr() as *mut x509_crt
+    }
+}
+
+
+#[repr(transparent)]
+pub struct MbedtlsList<T> {
+    inner: *mut T,
```
jethrogb, line 83:
```rust
inner: Option<MbedtlsBox<T>>
```

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub fn as_ref(&self) -> &Self {
+        self
+    }
+
+    pub fn empty(&self) -> bool {
+        self.inner == ::core::ptr::null_mut()
+    }
+    
+    pub fn push(&mut self, certificate: MbedtlsBox<Certificate>) -> () {
+        unsafe {
```
jethrogb, line 106:
```rust
self.append(MbedtlsList { inner: Some(certificate) })
```

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+pub struct Iter<'a> {
+    next: *const x509_crt,
+    r: PhantomData<&'a x509_crt>,
+}
+
+impl<'a> Iterator for Iter<'a> {
+    type Item = &'a Certificate;
```
jethrogb, line 181: Make `Item = &'a MbedtlsBox<Certificate>` (also for IterMut). I think you can then define the struct as follows:
```rust
pub struct Iter<'a> {
    next: Option<&'a MbedtlsBox<Certificate>>,
}
```

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+use core::ptr::drop_in_place;
+use core::marker::PhantomData;
+use crate::x509::Certificate;
+use crate::error::{Error, Result};
+use core::ptr::NonNull;
+
+#[repr(transparent)]
+pub struct MbedtlsBox<T> {
+    inner: NonNull<T>
+}
```
jethrogb, line 26: You can put the generic stuff (`MbedtlsBox<T>`, `MbedtlsList<T>`, `Drop`) in here (it should be up one level in the module hierarchy), but the rest should go in `certificate.rs`.

mbedtls/src/x509/mbedtls_wrappers.rs
```diff
@@ -0,0 +1,241 @@
...
+                (*cur).next = list.inner as *mut x509_crt;
+            }
+            core::mem::forget(list);
+        }
+    }
+
+    pub fn iter(&self) -> Iter<'_> {
+        Iter { next: self.inner as *const x509_crt, r: PhantomData }
+    }
+}
```
jethrogb, line 138: Add an `IntoIter` that deconstructs the list, giving you boxes.

mbedtls/src/x509/certificate.rs
```diff
@@ -153,16 +77,47 @@ fn x509_buf_to_vec(buf: &x509_buf) -> Vec<u8> {
...
+    pub fn from_pem(pem_or_der: &[u8]) -> Result<MbedtlsList<Certificate>> {
+        let cert = MbedtlsBox::<Certificate>::new()?;
+        unsafe { x509_crt_parse(cert.as_ref().into(), pem_or_der.as_ptr(), pem_or_der.len()) }.into_result()?;
+
+        let mut list = MbedtlsList::<Certificate>::new();
+        list.push(cert);
+        Ok(list)
+    }
+    
+    pub fn from_cert(cert: *const x509_crt) -> MbedtlsList::<Certificate> {
```
jethrogb, line 106: This doesn't need to be `pub`. Probably just inline it in `Clone::clone`?

mbedtls/src/x509/certificate.rs
```diff
@@ -153,16 +77,47 @@ fn x509_buf_to_vec(buf: &x509_buf) -> Vec<u8> {
...
-impl LinkedCertificate {
+
+impl Certificate {
+    pub fn from_der(der: &[u8]) -> Result<MbedtlsBox<Certificate>> {
+        let cert = MbedtlsBox::<Certificate>::new()?;
+        unsafe { x509_crt_parse_der(cert.as_ref().into(), der.as_ptr(), der.len()) }.into_result()?;
+        Ok(cert)
+    }
+
+    pub fn from_pem(pem_or_der: &[u8]) -> Result<MbedtlsList<Certificate>> {
```
jethrogb, line 97: `from_pem` should check that there is only one certificate and then add `from_pem_multiple` that returns a list

mbedtls/src/x509/mbedtls_wrappers.rs
AdrianCX: Option is an enum and is not transparent - https://doc.rust-lang.org/src/core/option.rs.html#158-167

We break all casting and logic by doing that.

mbedtls/src/x509/mbedtls_wrappers.rs
AdrianCX: Can't use option, switched to:

```
        self.append(MbedtlsList::<Certificate> { inner: certificate.inner.as_ptr() });
        core::mem::forget(certificate);
```

mbedtls/src/x509/mbedtls_wrappers.rs
AdrianCX: Since we need pointer to pointer in order to cast to the box it got a bit interesting, let me know if there are better options.

mbedtls/src/rng/hmac_drbg.rs
AdrianCX: This I would leave for a later PR if possible.

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +460,337 @@ impl<'a> Builder<'a> {
...
+        unsafe {
+            let mut p = self.next.take();
+            self.next = p.as_mut().map(|v| UnsafeFrom::from(&mut (v.inner.as_mut().inner.next) as *mut _ as *mut *mut x509_crt)).flatten();
+            p
+        }
+    }
+}
+
+/// Follows same pattern as IntoIter from rust-lang's linked_list.rs
+/// - line 108: https://github.com/rust-lang/rust/blob/master/library/alloc/src/collections/linked_list.rs
```
jethrogb, line 750: This should not be a doc comment

mbedtls/src/x509/certificate.rs
```diff
@@ -935,28 +1051,464 @@ cYp0bH/RcPTC0Z+ZaqSWMtfxRrk63MJQF9EXpDCdvQRcTMD9D85DJrMKn8aumq0M
...
+            assert_eq!(c1_info, format!("{:?}", it.next().unwrap()));
+            assert_eq!(c2_info, format!("{:?}", it.next().unwrap()));
+            assert_eq!(c3_info, format!("{:?}", it.next().unwrap()));
+            assert!(it.next().is_none());
+        }
+    }
+
+
+    #[test]
+    fn empty_certificate_test() {
```
jethrogb, line 1372: This test is not useful. All `Certificate`s should contain an actual certificate.

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +460,337 @@ impl<'a> Builder<'a> {
...
+    fn clone(&self) -> MbedtlsBox<Certificate> {
+        unsafe {
+            let len = (**self).inner.raw.len;
+            if len == 0 {
+                // We have a certificate that was never initialized.
+                // Panicking if memory failed to allocate since we don't have a Result<> on Clone.
+                MbedtlsBox::<Certificate>::new().unwrap()
+            } else {
+                // We have a certificate that was already created once, failing the re-parse is a possible memory corruption or bad usage of mbedtls internals.
+                Certificate::from_der(&std::slice::from_raw_parts((**self).inner.raw.p, len)).expect("Failed re-parsing existing DER")
```
jethrogb, line 499: Unnecessary `&`

mbedtls/src/x509/certificate.rs
AdrianCX: This is useful to verify we have null pointer checks everywhere needed and don't crash, otherwise I would need to find a der/pem that translates to an empty but parseable certificate.

Otherwise I agree, we cannot have a certificate that was never parsed.

mbedtls/src/rust_malloc.c
```diff
@@ -0,0 +1,29 @@
...
+#include MBEDTLS_CONFIG_FILE
+#endif
+
+#if defined(MBEDTLS_PLATFORM_C)
+#include "mbedtls/platform.h"
+#else
+#include <stdlib.h>
+#define mbedtls_calloc    calloc
+#define mbedtls_free      free
+#endif
```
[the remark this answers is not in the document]
AdrianCX: Intent is to follow what all mbedtls c files do so closing on assumption that's correct. let me know if I still need to change this.

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +460,316 @@ impl<'a> Builder<'a> {
...
+    pub fn iter(&self) -> Iter<'_> {
+        Iter { next: self.inner.as_ref(), r: PhantomData }
+    }
+
+    pub fn iter_mut(&mut self) -> IterMut<'_> {
+        IterMut { next: self.inner.as_mut(), r: PhantomData }
+    }
+}
+
+impl<> Clone for MbedtlsList<Certificate> {
```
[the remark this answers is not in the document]
AdrianCX: Added FromIterator similar to what I see in rust linked list.

That means changing the clone is now done via:
```
        self.iter().cloned().collect()
```

mbedtls/src/x509/certificate.rs
AdrianCX: re-open if you want me to remove the tests and I will

mbedtls/src/rng/hmac_drbg.rs
AdrianCX: Updated macro to allow members - this fixes hmac_drbg and context.

For the structures that have Box<c_struct> I have a few options - need to discuss with you which one is best.

AdrianCX: Added one of the options for the Box<> macro, I have a few other ideas if that doesn't look ok

mbedtls/src/ssl/context.rs
```diff
@@ -158,70 +181,37 @@ impl<'config> Context<'config> {
...
-    }
-}
-
-impl<'ctx> UnsafeFrom<*mut ssl_context> for HandshakeContext<'ctx> {
-    unsafe fn from(ctx: *mut ssl_context) -> Option<HandshakeContext<'ctx>> {
-        ctx.as_mut().map(|ctx| HandshakeContext { inner: ctx })
-    }
-}
-
-impl<'a> Session<'a> {
```
jethrogb, line 224: I prefer having the type-level separation. Here's one way to do it without having to duplicate the type:
```rust
pub enum Fresh {}
pub enum Established {}

pub struct Context<S = Fresh> { ... }
pub type Session = Context<Established>;

impl<S> Context<S> {
    pub fn new() -> Context { ... }
    pub fn establish(self) -> Result<Session, (Self, Error, Option<VerifyError>)> { ... }
}

impl Session {
    // functions that only make sense when established
}
```

You can go a step further and also have a `enum Handshake` and use that instead of the separate HandshakeContext type.

mbedtls/src/ssl/context.rs
```diff
@@ -72,65 +65,98 @@ impl<IO: Read + Write> IoCallback for IO {
...
+            handshake_pk: vec![],
+        }
     }
 
-    pub fn establish<'c, F: IoCallback>(
-        &'c mut self,
-        io: &'c mut F,
-        hostname: Option<&str>,
-    ) -> Result<Session<'c>> {
+    pub fn establish<T: IoCallback + Send + Sync + 'static>(&mut self, io: T, hostname: Option<&str>) -> Result<()> {
```
jethrogb, line 120: This should also include `VerifyError` (from verify_result) in the error result. Then verify_result should no longer be callable on a fresh context.

mbedtls/src/ssl/context.rs
```diff
@@ -251,62 +241,152 @@ impl<'a> Session<'a> {
...
-            unsafe { self.inner.session.as_ref().unwrap().ciphersuite as u16 }
+            return Err(Error::SslBadInputData);
         }
+        
+        Ok(unsafe { self.inner.session.as_ref().unwrap().ciphersuite as u16 })
     }
 
-    pub fn peer_cert(&self) -> Option<crate::x509::certificate::Iter> {
-        unsafe { UnsafeFrom::from(ssl_get_peer_cert(self.inner)) }
+    // This makes a copy of the certificate as peer_cert may be destroyed outside our control.
```
jethrogb, line 258: Stale comment?

mbedtls/src/ssl/context.rs
```diff
@@ -251,62 +241,152 @@ impl<'a> Session<'a> {
...
+    pub fn peer_cert(&self) -> Result<Option<&MbedtlsList<Certificate>>> {
+        if self.inner.session == ::core::ptr::null_mut() {
+            return Err(Error::SslBadInputData);
+        }
+
+        unsafe {
+            let peer_cert : &MbedtlsList<Certificate> = UnsafeFrom::from(&((*self.inner.session).peer_cert) as *const *mut x509_crt as *const *const x509_crt).ok_or(Error::SslBadInputData)?;
```
jethrogb, line 265: Add a comment explaining why you can't just call the function

mbedtls/src/ssl/context.rs
```diff
@@ -251,62 +241,152 @@ impl<'a> Session<'a> {
...
+}
 
-    pub fn verify_result(&self) -> StdResult<(), VerifyError> {
-        match unsafe { ssl_get_verify_result(self.inner) } {
-            0 => Ok(()),
-            flags => Err(VerifyError::from_bits_truncate(flags)),
+impl Drop for Context {
+    fn drop(&mut self) {
+        unsafe {
+            ssl_close_notify(self.into());
```
jethrogb, line 274: FYI: I'm guessing this may do IO? This may need revisiting once we implement async IO.

mbedtls/src/ssl/ticket.rs
```diff
@@ -20,46 +28,58 @@ pub trait TicketCallback {
...
 );
 
-impl<'rng> TicketContext<'rng> {
-    pub fn new<F: crate::rng::Random>(
-        rng: &'rng mut F,
+#[cfg(feature = "threading")]
+unsafe impl Sync for TicketContext {}
+
+impl TicketContext {
+    pub fn new<T: RngCallback + Send + Sync + 'static>(
```
jethrogb, line 59: Shouldn't need `+ Sync`

mbedtls/src/ssl/ticket.rs
```diff
@@ -20,46 +28,58 @@ pub trait TicketCallback {
...
-    fn data_ptr(&mut self) -> *mut c_void;
+    fn data_ptr(&self) -> *mut c_void;
 }
 
+
 define!(
-    #[c_ty(ssl_ticket_context)]
-    struct TicketContext<'rng>;
-    const init: fn() -> Self = ssl_ticket_init;
+    #[c_box_ty(ssl_ticket_context)]
```
jethrogb, line 44: Please point to a specific reason boxing is needed here

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +464,287 @@ impl<'a> Builder<'a> {
...
+impl Clone for MbedtlsBox<Certificate> {
+    fn clone(&self) -> MbedtlsBox<Certificate> {
+        unsafe {
+            let len = (**self).inner.raw.len;
+            if len == 0 {
+                // We have a certificate that was never initialized.
+                //
+                // This is not possible from 'mbedtls' side as all certificates are first init then they parse a der/pem.
+                // This is also not possible from rust side as all init() calls are followed by parsing.
+                panic!("Trying to clone an uninitialized certificate.");
```
jethrogb, line 510: Actually, use the `unreachable!` macro to better signal intent

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +464,287 @@ impl<'a> Builder<'a> {
...
+
+impl<'a> UnsafeFrom<*mut *mut x509_crt> for &'a mut MbedtlsBox<Certificate> {
+    unsafe fn from(ptr: *mut *mut x509_crt) -> Option<&'a mut MbedtlsBox<Certificate>> {
+        if ptr.is_null() || (*ptr).is_null() {
+            return None;
+        }
+
+        (ptr as *mut MbedtlsBox<Certificate>).as_mut()
+    }
+}
```
jethrogb, line 549: Unused, delete.

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +464,287 @@ impl<'a> Builder<'a> {
...
+    }
+
+    pub(crate) unsafe fn inner_ffi_mut(&self) -> *mut x509_crt {
+        self.inner.as_ref().map_or(::core::ptr::null_mut(), |c| c.inner.as_ptr() as *mut x509_crt)
+    }
+}
+
+impl Clone for MbedtlsList<Certificate> {
+    fn clone(&self) -> MbedtlsList<Certificate> {
+        self.iter().cloned().collect()
```
jethrogb, line 616: Nice.

mbedtls/src/x509/certificate.rs
```diff
@@ -676,6 +460,337 @@ impl<'a> Builder<'a> {
...
+}
+
+impl<'a> UnsafeFrom<*const *const x509_crt> for &'a MbedtlsList<Certificate> {
+    unsafe fn from(ptr: *const *const x509_crt) -> Option<&'a MbedtlsList<Certificate>> {
+        if ptr == ::core::ptr::null_mut() {
+            return None;
+        }
+        if *ptr == ::core::ptr::null_mut() {
+            return None;
+        }
```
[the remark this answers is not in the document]
jethrogb: This is not resolved

mbedtls/valgrind_unittests.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
jethrogb, line 1: Can this be run as part of CI?

jethrogb (commented):
I think this only partially addresses #9, please update the PR description so it doesn't close the issue on merge.

mbedtls/src/ssl/context.rs
```diff
@@ -6,42 +6,39 @@
...
-        user_data: *mut c_void,
-        data: *const c_uchar,
-        len: size_t,
-    ) -> c_int;
 
+pub trait IoCallback {
+    unsafe extern "C" fn call_recv(user_data: *mut c_void, data: *mut c_uchar, len: size_t) -> c_int where Self: Sized;
+    unsafe extern "C" fn call_send(user_data: *mut c_void, data: *const c_uchar, len: size_t) -> c_int where Self: Sized;
     fn data_ptr(&mut self) -> *mut c_void;
+    fn as_any(&mut self) -> &mut dyn Any;
```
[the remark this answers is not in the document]
AdrianCX: That would make it harder to handle the Context<T> types.

I use this trick as I can store the IO as dyn pointer and avoid templates on Context.
```
        io: Option<Box<dyn IoCallback>>,
```

This allows easier handling of Context class as I don't need to know what T is, only the callers care when they call:
```
    pub fn get_mut_io<T: IoCallback + 'static>(&mut self) -> IoResult<&mut T> {
```

Easier handling means easy casting to Context when callback from mbedtls is triggered among other things.

mbedtls/src/ssl/context.rs
AdrianCX: Why this may be difficult to use / understand:

- We cannot have multiple 'Session' per Context.

- Any new session overrides/closes the old one in the Context.

- Context may manipulate Session indirectly.

As such - I just put everything in Context to avoid any logic issues.

I could make the Session a reference to Context with lifetime but I wanted to avoid lifetimes...

I also wanted to avoid making Context generic due to reasons in another comment above.

mbedtls/src/lib.rs
```diff
@@ -163,3 +165,11 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut time_t) -> time_t {
     }
     timestamp
 }
+
+// Debug not available in SGX
```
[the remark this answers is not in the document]
AdrianCX: SGX does not define the MBEDTLS_DEBUG_C needed to have function available.

mbedtls-sys/build/config.rs:408
```
    ("std",                   ("MBEDTLS_DEBUG_C",                           Defined)),
```

But this is ignored due to: mbedtls-sys/build/build.rs:43
```
        for &(feat, def) in config::FEATURE_DEFINES {
            if (feat == "std") && (target == "x86_64-fortanix-unknown-sgx") {
                continue;
            }
```

So I'm assuming there was a reason to not have it available. (probably because keys can leak via debug logs if you set it to verbose).

Let me know if I should change this.

mbedtls/src/rng/os_entropy.rs
```diff
@@ -6,71 +6,107 @@
...
-callback!(EntropySourceCallback(data: *mut c_uchar, size: size_t, out: *mut size_t) -> c_int);
+callback!(EntropySourceCallbackMut,EntropySourceCallback(data: *mut c_uchar, size: size_t, out: *mut size_t) -> c_int);
 
 define!(
-    #[c_ty(entropy_context)]
-    struct OsEntropy<'source>;
-    pub const new: fn() -> Self = entropy_init;
+    // Moving data causes dangling pointers: https://github.com/ARMmbed/mbedtls/issues/2147
+    // Storing data in heap and forcing rust move to only move the pointer (box) referencing it.
+    // The move will be faster. Access to data will be slower due to additional indirection.
```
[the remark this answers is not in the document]
AdrianCX: I was a bit overzealous with moving to boxes, this is safe without a box.

mbedtls/valgrind_unittests.sh
AdrianCX: It would need something to interpret the valgrind logs.

That is something I was planning on doing with another PR, let me know if I should raise an issue for that.

I can also take the time to do it now if urgent.

mbedtls/src/ssl/context.rs
AdrianCX: correct, it can do IO.

mbedtls/src/ssl/context.rs
```diff
@@ -251,62 +241,152 @@ impl<'a> Session<'a> {
...
+        self.context.handshake_crl = crl;
+        Ok(())
+    }
+
+    /// If this is never called, will use the set of private keys and
+    /// certificates configured in the `Config` associated with this `Context`.
+    /// If this is called at least once, all those are ignored and the set
+    /// specified using this function is used.
+    ///
+    /// @warning: chain/key must live the same as the SNI callback otherwise risk of dangling pointers.
```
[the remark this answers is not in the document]
AdrianCX: The warning is stale. It was before I added Arcs to hold these certificates as well.

Removed the warning

mbedtls/src/ssl/context.rs
```diff
@@ -72,65 +65,98 @@ impl<IO: Read + Write> IoCallback for IO {
...
+        self.handshake_cert.clear();
+        self.handshake_pk.clear();
+        self.handshake_ca_cert = None;
+        self.handshake_crl = None;
+
         unsafe {
-            ssl_session_reset(&mut self.inner).into_result()?;
+            let mut io = Box::new(io);
+            
+            ssl_session_reset(self.into()).into_result()?;
```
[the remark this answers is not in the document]
AdrianCX: thank you!

mbedtls/src/ssl/context.rs
```diff
@@ -251,62 +241,152 @@ impl<'a> Session<'a> {
...
+/// That is due to SNI being held by an Arc inside Config.
+/// Config lives longer then Context. Context lives longer then Handshake.
+///
+/// Alternatives are not possible due to:
+/// - mbedtls not providing any callbacks on handshake finish.
+/// - no reasonable way to obtain a storage within the sni callback tied to the handshake or to the rust Context. (without resorting to a unscalable map or pointer magic that mbedtls may invalidate)
+///
+impl<'ctx> HandshakeContext<'ctx> {
+
+    pub fn init(context: &'ctx mut Context) -> Self {
```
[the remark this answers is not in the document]
AdrianCX: Used by mbedtls/src/ssl/config.rs:567

```
    let mut ctx = HandshakeContext::init(context);
```

mbedtls/src/rng/os_entropy.rs
```diff
@@ -6,71 +6,107 @@
...
-                    ENTROPY_SOURCE_STRONG
-                } else {
-                    ENTROPY_SOURCE_WEAK
-                }
+                if strong { ENTROPY_SOURCE_STRONG } else { ENTROPY_SOURCE_WEAK }
             )
             .into_result()?
         };
+
+        // Rust ensures only one mutable reference is currently in use.
```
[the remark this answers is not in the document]
AdrianCX: We can discuss an example - I can add that as a test - then I can start putting in Mutex<Vec<_>>

The usage is a bit unclear for the concurrent use case, I'm assuming you're thinking of passing os_entropy to Contexts and independently updating it via 'update_manual' from rust side on a separate thread?

mbedtls/src/ssl/ticket.rs
AdrianCX: reverted to non-box.

mbedtls/src/ssl/context.rs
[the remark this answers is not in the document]
AdrianCX: made it a pub(crate), please re-open if not acceptable.

mbedtls/src/ssl/config.rs
```diff
@@ -83,57 +93,148 @@ define!(
...
 
 define!(
-    #[c_ty(ssl_config)]
-    struct Config<'c>;
-    const init: fn() -> Self = ssl_config_init;
-    const drop: fn(&mut Self) = ssl_config_free;
-    impl<'q> Into<ptr> {}
-    impl<'q> UnsafeFrom<ptr> {}
+    // Moving data may cause dangling pointers: https://github.com/ARMmbed/mbedtls/issues/2147
+    // Storing data in heap and forcing rust move to only move the pointer (box) referencing it.
```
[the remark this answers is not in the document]
AdrianCX: switched back to non-box

mbedtls/src/rng/ctr_drbg.rs
```diff
@@ -6,126 +6,82 @@
...
-            CtrDrbgInner {
-                inner,
-                r: PhantomData,
-            }
-        }
-    }
+    // If threading enabled - we are Sync, but for that to be true we need entropy to also be sync.
+    #[cfg(feature = "threading")]
+    pub fn new<T: EntropyCallback + Send + Sync + 'static>(entropy: Arc<T>, additional_entropy: Option<&[u8]>) -> Result<Self> {
+        let mut inner = Box::new(ctr_drbg_context::default());
```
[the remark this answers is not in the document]
AdrianCX: I added some code for default members.

If that's not sufficient we need to pass parameters to init() funciton and I can raise a ticket. (let me know if so)

mbedtls/src/ssl/context.rs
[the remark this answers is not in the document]
jethrogb: Thanks for pointing out we can setup static dispatch for the most part before converting to a trait object. That's good. But instead of adding this method to this trait, I think you should do this:

```
trait IoCallback: Any { ... }

struct Context {
    io: Option<Box<dyn Any>>,
    ...
}

impl Context {
    fn io(&self) -> &dyn Any { ... }
    fn io_mut(&mut self) -> &mut dyn Any { ... }
    // possibly also add:
    fn into_io(self) -> Box<dyn Any> { ... }
}
```

mbedtls/src/ssl/config.rs
```diff
@@ -155,186 +256,165 @@ impl<'c> Config<'c> {
...
 }
 
 /// Builds a linked list of x509_crt instances, all of which are owned by mbedtls. That is, the
 /// memory for these certificates has been allocated by mbedtls, on the C heap. This is needed for
 /// situations in which an mbedtls function takes ownership of a list of certs. The problem with
-/// handing such functions a "normal" cert list such as certificate::LinkedCertificate or
+/// handing such functions a "normal" cert list such as certificate::Certificate or
 /// certificate::List, is that those lists (at least partly) consist of memory allocated on the
 /// rust-side and hence cannot be freed on the c-side.
 pub struct ForeignOwnedCertListBuilder {
```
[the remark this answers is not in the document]
AdrianCX: nice, thanks

mbedtls/src/ssl/context.rs
AdrianCX: Parking for now until we get native-tls compat running

mbedtls/src/ssl/config.rs
```diff
@@ -288,125 +379,80 @@ impl<'c> Config<'c> {
...
             let cb = &mut *(closure as *mut F);
-            let child: &LinkedCertificate = UnsafeFrom::from(child).expect("valid child certificate");
-            let mut cert_builder = ForeignOwnedCertListBuilder::new();
-            match cb(child, &mut cert_builder) {
-                Ok(()) => {
-                    *candidate_cas = cert_builder.to_x509_crt_ptr();
+            let crt: &MbedtlsList<Certificate> = UnsafeFrom::from(&child as *const *const x509_crt).expect("valid certificate");
+            match cb(&crt) {
+                Ok(list) => {
+                    *candidate_cas = list.leak();
```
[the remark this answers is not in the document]
AdrianCX: It's not leaked, you are correct, I'll put investigation below.

the 'leak' name comes from rust Box leak: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak

I'll rename to into_raw.

Leak check (I will put in comment as well): mbedtls-sys/vendor/library/x509_crt.c:2904

Code first frees previous CA Callback returned data before fetching new data via 'f_ca_cb'
```
            mbedtls_x509_crt_free( ver_chain->trust_ca_cb_result );
            mbedtls_free( ver_chain->trust_ca_cb_result );
            ver_chain->trust_ca_cb_result = NULL;

            ret = f_ca_cb( p_ca_cb, child, &ver_chain->trust_ca_cb_result );
            if( ret != 0 )
                return( MBEDTLS_ERR_X509_FATAL_ERROR );

            cur_trust_ca = ver_chain->trust_ca_cb_result;
```

mbedtls/src/ssl/context.rs
```diff
@@ -6,42 +6,38 @@
...
     fn data_ptr(&mut self) -> *mut c_void;
 }
 
-impl<IO: Read + Write> IoCallback for IO {
-    unsafe extern "C" fn call_recv(
-        user_data: *mut c_void,
-        data: *mut c_uchar,
-        len: size_t,
-    ) -> c_int {
+impl<IO: Read + Write + 'static> IoCallback for IO {
```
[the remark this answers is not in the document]
AdrianCX: I'll add the 'static bound to other callbacks as well. I can't have members hold references.

mbedtls/valgrind_unittests.sh
jethrogb: Ok please open a GH issue then

mbedtls/src/rng/ctr_drbg.rs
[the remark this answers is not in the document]
AdrianCX: https://github.com/fortanix/rust-mbedtls/issues/131

mbedtls/src/rng/os_entropy.rs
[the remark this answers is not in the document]
AdrianCX: removed mutex

mbedtls/src/ssl/config.rs
```diff
@@ -155,112 +226,127 @@ impl<'c> Config<'c> {
...
+            // mbedtls-sys/vendor/library/ssl_srv.c - mbedtls_ssl_handshake_server_step
+            // mbedtls-sys/vendor/library/ssl_srv.c - ssl_parse_client_hello
+            // mbedtls-sys/vendor/library/ssl_srv.c - ssl_parse_servername_ext
+            //
+            // As such:
+            // - The ssl_context is a rust 'Context' structure that we have a mutable reference to via 'establish'
+            // - We can pointer cast to it to allow storing additional objects.
+            //
+            if closure.is_null() || ctx.is_null() || name.is_null() {
+                return ::mbedtls_sys::ERR_X509_BAD_INPUT_DATA;
```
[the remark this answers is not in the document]
AdrianCX: Looking at mbedtls code at the moment these can't be null:

mbedtls-sys/vendor/library/ssl_srv.c:120
```
            ret = ssl->conf->f_sni( ssl->conf->p_sni,
                                    ssl, p + 3, hostname_len );
```

Let me know if I should remove. These are usually useful to catch software errors and not crash / corrupt.

mbedtls/valgrind_unittests.sh
AdrianCX: https://github.com/fortanix/rust-mbedtls/issues/132

mbedtls/src/ssl/context.rs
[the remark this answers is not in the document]
AdrianCX: This is needed due to IoCallback deriving from Any now.

Adding this:
```
use std::any::Any;

pub trait IoCallback : Any {
```

Results in:
```
error[E0310]: the parameter type `IO` may not live long enough
  --> mbedtls/src/ssl/context.rs:41:24
   |
41 | impl<IO: Read + Write> IoCallback for IO {
   |      ---               ^^^^^^^^^^ ...so that the type `IO` will meet its required lifetime bounds
   |      |
   |      help: consider adding an explicit lifetime bound...: `IO: 'static +`
```

mbedtls/src/pk/dhparam.rs
```diff
@@ -22,6 +22,7 @@ impl Dhm {
     /// Takes both DER and PEM forms of FFDH parameters in `DHParams` format.
     ///
     /// When calling on PEM-encoded data, `params` must be NULL-terminated
+    #[allow(dead_code)]
```
jethrogb, line 25: Why is this dead code? This is called from ssl::Config?

jethrogb: Actually, you should make this `pub` now that you no longer call it from ssl::Config

jethrogb:
bors try

PR #130, opened 2020-12-12
Add DSA signatures

randombit:
~WRT side channels I assumed that `mod_exp`, `modulo` etc are constant time. I have not verified if this is true for current version of mbedtls. May be worth adding blinding.~ Added blinding.


mbedtls/src/dsa/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+    Ok(c)
+}
+
+fn encode_dsa_signature(q_bits: usize, r: &Mpi, s: &Mpi) -> Result<Vec<u8>> {
+    let q_bytes = (q_bits + 7) / 8;
+    let r = r.to_binary_padded(q_bytes)?;
+    let s = s.to_binary_padded(q_bytes)?;
+    let mut sig = Vec::with_capacity(r.len() + s.len());
+    sig.extend_from_slice(&r);
+    sig.extend_from_slice(&s);
```
jethrogb, line 215: Avoid copy:
```
    let mut sig = r;
    sig.extend_from_slice(&s);
```

mbedtls/src/dsa/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+    params: DsaParams,
+    y: Mpi,
+}
+
+const DSA_OBJECT_IDENTIFIER: &[u64] = &[1, 2, 840, 10040, 4, 1];
+
+impl DsaPublicKey {
+
+    pub fn from_params(params: DsaParams, y: Mpi) -> Result<Self> {
+        Ok(Self { params, y })
```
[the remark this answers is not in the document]
randombit: Yeah we can check that it is within range of `[1..p)` and also that it is of order `q`.

mbedtls/src/dsa/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+    let bits = upper_bound.bit_length()?;
+    let mut rnd_buf = vec![0u8; (bits + 7 + 64) / 8];
+    rng.random(&mut rnd_buf)?;
+    let c = Mpi::from_binary(&rnd_buf)?;
+    let mut c = c.modulo(&(upper_bound - 1)?)?;
+    c += 1;
+    Ok(c)
+}
+
+fn encode_dsa_signature(q_bits: usize, r: &Mpi, s: &Mpi) -> Result<Vec<u8>> {
```
jethrogb, line 209: On second thought, for ECDSA, we use DER-encoded signatures. Is this specified the same way for DSA? We should probably use that.

randombit: Yeah for DSA and ECDSA both methods are about equally common depending on the standard. DSA in X.509 uses DER IIRC.

mbedtls/src/dsa/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+        if signature.len() % 2 == 1 {
+            return Ok(false);
+        }
+
+        let p = &self.params.p;
+        let q = &self.params.q;
+
+        let half = signature.len() / 2;
+        let r = Mpi::from_binary(&signature[..half])?;
+        let s = Mpi::from_binary(&signature[half..])?;
```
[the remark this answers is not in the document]
randombit: moot wrt changing to der signatures

mbedtls/src/dsa/mod.rs
randombit: moot wrt changing to der signatures

mbedtls/src/dsa/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+        let zero = Mpi::new(0)?;
+        if r == zero || s == zero {
+            return Err(Error::MpiBadInputData);
+        }
+        encode_dsa_signature(q.bit_length()?, &r, &s)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
randombit: Done for public key. Currently for private keys we are using PKCS8 form of DSA. Do you want me to change this to instead be OpenSSL's format for private keys? Difference is

`openssl gendsa dsaparams.pem` produces something which is basically OpenSSL specific (afaik)
`openssl pkcs8 -topk8 -nocrypt -in dsapriv.pem` produces a PKCS8 which is what everything else expects

mbedtls/src/dsa/mod.rs
randombit: Done

mbedtls/src/dsa/mod.rs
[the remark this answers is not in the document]
jethrogb: We want PKCS8

mbedtls/Cargo.toml
```diff
@@ -52,7 +54,7 @@ cc = "1.0"
 
 [features]
 # Features are documented in the README
-default = ["std", "aesni", "time", "padlock"]
+default = ["std", "aesni", "time", "padlock", "dsa"]
```
jethrogb, line 57: Let's not enable this by default. Also update README

mbedtls/src/pk/mod.rs
```diff
@@ -23,12 +23,15 @@ use byteorder::{BigEndian, ByteOrder};
 
 pub(crate) mod dhparam;
 mod ec;
-mod rfc6979;
+pub(crate) mod rfc6979;
```
jethrogb, line 26: No longer needed after module move

mbedtls/src/pk/dsa/mod.rs
```diff
@@ -0,0 +1,592 @@
...
+        let md: MdInfo = match md_type.into() {
+            Some(md) => md,
+            None => panic!("no such digest"),
+        };
+        let rfc6979_nonce = generate_rfc6979_nonce(&md, &self.x, &self.params.q, pre_hashed_message)?;
+        let k = Mpi::from_binary(&rfc6979_nonce)?;
+        self.sign_with_explicit_nonce(pre_hashed_message, k, rng)
+    }
+
+    // Exposed for testing
```
jethrogb, line 316: Stale comment

jethrogb (commented):
Mostly good, just waiting for the private key DER test

mbedtls/Cargo.toml
randombit: OK. I added `cargo test --features dsa` to `ct.sh` so tests still run in CI.

PR #133, opened 2020-12-22
Upgrade to mbedtls 2.25.0

raoulstrackx:
Upgrades the `mbedtls-sys` crate to version 2.25.0.

*Warning* The upstream mbedtls library deprecated the functions `mbedtls_cipher_auth_encrypt` and `mbedtls_cipher_auth_decrypt` to `mbedtls_cipher_auth_encrypt_ext` and and `mbedtls_cipher_auth_decrypt_ext` respectively. The new functions assume that the tag is appended to the ciphertext. The current PR exposes a similar requirement. This saves explicit memory allocation and copy operations, at the cost of a changed API.
 
See https://github.com/ARMmbed/mbedtls/blob/development/include/mbedtls/cipher.h#L1027

cc: @jethrogb


mbedtls-sys/build/config.rs
```diff
@@ -66,6 +66,7 @@ for line in open('vendor/include/mbedtls/config.h').readlines():
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 const DEFAULT_DEFINES: &'static [CDefine] = &[
+    ("MBEDTLS_CONFIG_H",                                  Defined),
```
jethrogb, line 69: This should not be defined manually, it's already included in `PREFIX` above.

mbedtls-sys/build/config.rs
```diff
@@ -382,7 +388,7 @@ const DEFAULT_DEFINES: &'static [CDefine] = &[
     ("MBEDTLS_SSL_DEFAULT_TICKET_LIFETIME",               Undefined), // default: 86400
     ("MBEDTLS_PSK_MAX_LEN",                               Undefined), // default: 32
     ("MBEDTLS_SSL_COOKIE_TIMEOUT",                        Undefined), // default: 60
-    ("MBEDTLS_SSL_CIPHERSUITES",                          Undefined), // no default
```
jethrogb, line 385: I think I mentioned this before, the default here refers to what you get when you don't define it. That's different from what's shown in the commented-out define in the config.h source code.

mbedtls/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mbedtls"
-version = "0.6.1"
+version = "0.7.0"
```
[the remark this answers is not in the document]
raoulstrackx: Addressed 1, will talk about 2 offline

PR #135, opened 2020-12-23
Clean up features

jethrogb:
Features are now split into two categories: cargo features and
platform features. Platform features are dependent on the
compilation target and automatically selected when needed. This
removes the (broken) mechanism of selecting backend implementations
based on cargo features. Enabled platform features are passed from
`mbedtls-sys` to dependent crates so they can change their
implementation as needed.

As part of this change, it's no longer possible to build the
`mbedtls` crate without threading support. Support may be added
again in the future if there's a need for it. The `mbedtls-sys`
crate can still be built without threading support.

The changes to `mbedtls-sys` are intended to not be a breaking
change (except for the removal of the libc feature, which had
not been explicitly advertised to users).

Also some cleanup related to Rust 2018 edition.

### DIFF
--- a/mbedtls/tests/hyper.rs
+++ b/mbedtls/tests/hyper.rs
@@ -174,9 +174,6 @@ mod tests {
     use mbedtls::ssl::TicketContext;
     
     #[cfg(not(target_env = "sgx"))]
-    use mbedtls::set_global_debug_threshold;
-    
-    #[cfg(not(target_env = "sgx"))]
     use mbedtls::rng::{OsEntropy, CtrDrbg};
 
     #[cfg(target_env = "sgx")]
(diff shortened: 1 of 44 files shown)


mbedtls/tests/hyper.rs
```diff
@@ -283,9 +280,6 @@ mod tests {
         };
         config.set_dbg_callback(dbg_callback);
 
-        #[cfg(not(target_env = "sgx"))]
-        unsafe { set_global_debug_threshold(1); }
```
AdrianCX, line 287: please keep this if possible as it's been very useful debugging code when unit test fails.

AdrianCX (approved):
I would keep the debug logs in hyper.rs to help other people troubleshoot issues. (has been useful for me)

Other then that it looks great and simplifies things a lot.

mbedtls/tests/hyper.rs
jethrogb: That would require limiting this test to run only when a certain feature is set, limiting its usefulness.

jethrogb:
bors r=AdrianCX

PR #141, opened 2021-01-08
Backport uninitialized memory fix #124

jseyfried:
Original description:

I noticed that the `decrypt()` function would take its return value from uninitialized memory when performing a raw RSA decrypt. Also, the length of the plaintext output buffer was not checked.

Edit: Some further changes were incorporated into this PR. The total list of changes is:
 * Fix returning uninitialized value from `Pk::decrypt()` when using RSA RAW mode
 * Remove unnecessary use of `MaybeUninit`
 * Add output buffer length check to `Pk::sign_deterministic()`

Co-authored-by: Pieter Agten <pieter.agten@fortanix.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,7 +235,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "mbedtls"
-version = "0.6.2"
+version = "0.6.3"
 dependencies = [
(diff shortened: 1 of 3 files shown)


mzohreva:
@jseyfried I messed up this PR. It shows as merged but it's not since after merge I forced pushed to v0.6 branch to erase the manual merge. Could you please open another PR with the same parameters?

PR #149, opened 2021-02-07
Add a quick test for decrypting openssl files

AdrianCX:
A quick example how to decrypt openssl files (and how to encrypt them if you follow bash script)

Adding this as a test so that we don't break anything going forward.


jethrogb:
What is this testing that we weren't already testing before? We already test pbkdf2 in tests/pbkdf.rs and various forms of CBC in tests/save_restore.rs.

AdrianCX:
maybe it's useful as example then?, how to encrypt/decrypt to/from openssl's utility.

I agree that the code individually was tested properly.

Otherwise I'll drop the review.

Taowyoo:
Close this because long time no updates

Issue #121, opened 2020-07-28
Build fails with bindgen errors


connorcl:
I've been trying to use this library in a `no_std` project, but the build of `mbedtls-sys-auto` always fails with a large number of Bindgen ERROR: unsupported type \`CXTypeKind(119)\` errors. I tried to make sure it wasn't the result of cross-compilation or a wrong selection of features by trying to include `mbedtls` as a dependency in an otherwise empty project (not `no_std`), but the build still fails with this same bindgen error.

jethrogb:
Duplicate of #5

jethrogb:
Final PR: #152

jethrogb:
Fixed in #152

Issue #61, opened 2019-10-07
Bindgen ERROR: unsupported type `CXTypeKind(119)`


MabezDev:
```
$ clang --version

clang version 8.0.1 (tags/RELEASE_801/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

Is there a specific version of clang required to build this project?

jethrogb:
Unfortunately the version of bindgen used requires clang 3.8 or older right now. We're working on fixing this.

Closing as duplicate of #5.

jethrogb:
Final PR: #152

jethrogb:
Fixed in #152

PR #152, opened 2021-03-11
Update bindgen

### LINKED ISSUE #5: Upgrade bindgen

jethrogb:
I took some time this week to make the necessary bindgen changes (https://github.com/rust-lang/rust-bindgen/pull/2004 https://github.com/rust-lang/rust-bindgen/pull/2006 https://github.com/rust-lang/rust-bindgen/pull/2007). This PR updates the bindgen build to use that.

Breaking changes:
* unnamed types are renamed
* C-unions are now actual unions
  * [x] the field accessor functions that used to exist are easy enough to add back
* bitfields are done differently now
* some fn-ptrs are now unsafe
* some fns and fn-ptrs now take const pointers as arguments
* havege.c and timing.c are now not compiled on non-unix platforms, i.e. SGX (probably wasn't working properly anyway)

Fixes #5 #14 #61 #72 #88 #121

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,22 +11,26 @@ os: linux
 dist: xenial
 addons:
   apt:
+    sources:
+    - sourceline: "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-11 main"
+      key_url: "https://apt.llvm.org/llvm-snapshot.gpg.key"
     packages:
-    - llvm-3.8-dev
-    - libclang-3.8-dev
-    - clang-3.8
+    - llvm-3.9-dev
+    - libclang-3.9-dev
+    - clang-3.9
+    - clang-11
     - cmake
 rust:
    - stable
    - beta
    - nightly
-   - nightly-2019-07-01
+   - nightly-2021-03-25
 env:
   global:
     - RUST_BACKTRACE=1
     # Pinned to this particular nightly version because of core_io. This can be
     # re-pinned whenever core_io is updated to the latest nightly.
-    - CORE_IO_NIGHTLY=nightly-2019-07-01
-    - LLVM_CONFIG_PATH=llvm-config-3.8
+    - CORE_IO_NIGHTLY=nightly-2021-03-25
+    - LLVM_CONFIG_PATH=llvm-config-3.9
 script:
   - ./ct.sh
--- a/mbedtls-sys/Cargo.toml
+++ b/mbedtls-sys/Cargo.toml
@@ -27,9 +27,12 @@ libc = { version = "0.2.0", optional = true }
 libc = { version = "0.2.0" }
 
 [build-dependencies]
-bindgen = "0.19.0"
+bindgen = "0.58.0"
 cmake = "0.1.17"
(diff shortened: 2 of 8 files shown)


mbedtls-sys/Cargo.toml
```diff
@@ -27,7 +27,7 @@ libc = { version = "0.2.0", optional = true }
 libc = { version = "0.2.0" }
 
 [build-dependencies]
-bindgen = "0.19.0"
+bindgen = { git = "https://github.com/jethrogb/rust-bindgen/", branch = "build" }
```
jethrogb, line 30: Need a bindgen release before this can be merged

.travis.yml
```diff
@@ -6,15 +6,16 @@ branches:
     - trying
     # Not really necessary, just to get a green badge on “master”
     - master
+    - acruceru/fix-cert-issue
```
jethrogb, line 9: This needs to be reverted (and the PR target branch changed) before merge

mbedtls-sys/Cargo.toml
```diff
@@ -27,9 +27,12 @@ libc = { version = "0.2.0", optional = true }
 libc = { version = "0.2.0" }
 
 [build-dependencies]
-bindgen = "0.19.0"
+bindgen = { git = "https://github.com/rust-lang/rust-bindgen/", branch = "master" }
```
raoulstrackx, line 30: I guess this will be updated to the new bindgen release before merging?

jethrogb: Yes, as mentioned above.

jethrogb:
The new version of bindgen depends on libloading 0.7. In 0.6, libloading introduced a `#[nonexhaustive]` enum that requires a newer rustc than the current CORE_IO_NIGHTLY. Options:

* Remove core_io support
* Update core_io
* Hack around libloading with a crate patch

jethrogb:
This is ready for review now

mbedtls-sys/build/bindgen.rs
```diff
@@ -8,78 +8,136 @@
...
+            .allowlist_var("^(?i)mbedtls_.*")
+            .allowlist_recursively(false)
+            .blocklist_type("^mbedtls_time_t$")
+            .use_core()
+            .ctypes_prefix("::types::raw_types")
+            .parse_callbacks(Box::new(MbedtlsParseCallbacks))
+            .default_enum_style(bindgen::EnumVariation::Consts)
+            .generate_comments(false)
+            .derive_copy(true)
+            .derive_debug(false) // used to be buggy, may be possible now?
```
raoulstrackx, line 119: Seems to work for `x86_64-fortanix-unknown-sgx` but for `x86_64-unknown-linux-gnu` it results in:
`libc::pthread_mutex_t` cannot be formatted using `{:?}` because it doesn't implement `Debug`. I would remove the comment.

PR #153, opened 2021-03-12
Fix cert expired in unit tests

AdrianCX:
Quick fix to unblock other PRs.

Will investigate why it hangs and raise another PR afterwards.
This solves the for now.


jethrogb:
`nocapture` shouldn't be necessary, failing tests will have their output printed anyway. If you want to enforce a timeout, please use https://crates.io/crates/ntest

PR #154, opened 2021-03-24
SGX compile options

raoulstrackx:
In order to compile for the SGX target, we want to be able to specify our own (e.g., clang) compiler and compile flags enabling LVI mitigations.


mbedtls-sys/build/cmake.rs
```diff
@@ -8,23 +8,33 @@
...
         cmk.cflag(format!(
             r#"-DMBEDTLS_CONFIG_FILE="\"{}\"""#,
             self.config_h.to_str().expect("config.h UTF-8 error")
         ))
         .define("ENABLE_PROGRAMS", "OFF")
         .define("ENABLE_TESTING", "OFF")
         .build_target("lib");
+        if env::var("CC").is_ok() {
+            cmk.define("CMAKE_C_COMPILER_FORCED", "TRUE");
+        }
```
jethrogb, line 28: Why is this needed?

mbedtls-sys/build/cmake.rs
```diff
@@ -8,23 +8,33 @@
...
         if FEATURES.have_platform_component("c_compiler", "freestanding") {
             cmk.cflag("-fno-builtin")
                 .cflag("-D_FORTIFY_SOURCE=0")
                 .cflag("-fno-stack-protector");
         }
+        if utils::env_have_target_cfg("env", "sgx") {
+            // Compile for sgx target is similar as for unix
+            cmk.cflag("-Dunix");
+        }
```
jethrogb, line 37: See #152 for the real solution

PR #157, opened 2021-04-08
Update mbedtls-sys-auto dependency

raoulstrackx:
The current `mbedtls-sys-auto` dependency won't allow us to upgrade to version 2.26.0.

cc: @jethrogb


jethrogb:
This PR doesn't make sense to me.

PR #158, opened 2021-04-09
Add sysroot support for GNU based toolchains

landhb:
This PR adds sysroot detection for GNU based toolchains for targets like `aarch64-unknown-linux-musl`. This prevents bindgen from using incorrect headers when generating bindings. 

For instance when using cross, the aarch64 musl sysroot is installed in /usr/local/aarch64-linux-musl/ but bindgen/clang will attempt to look for headers in /usr/include unless `--sysroot` is provided.

### DIFF
--- a/mbedtls-sys/build/build.rs
+++ b/mbedtls-sys/build/build.rs
@@ -106,19 +106,17 @@ impl BuildConfig {
         // Determine the sysroot for this compiler so that bindgen
         // uses the correct headers
         let compiler = cc::Build::new().get_compiler();
-        match compiler.is_like_gnu() {
-            true => {
-                let output = compiler.to_command().args(&["--print-sysroot"]).output().unwrap();
-                let output = std::str::from_utf8(&output.stdout).unwrap();
-                let res = output
+        let output = compiler.to_command().args(&["--print-sysroot"]).output();
+        match (output, compiler.is_like_gnu()) {
+            (Ok(sysroot), true) => {
+                let path = std::str::from_utf8(&sysroot.stdout).expect("Malformed sysroot");
+                let trimmed_path = path
                     .strip_suffix("\r\n")
-                    .or(output.strip_suffix("\n"))
-                    .unwrap_or(&output);
-                let mut val = "--sysroot=".to_string();
-                val.push_str(&res);
-                cflags.push(val)
-            },
-            false => {}
+                    .or(path.strip_suffix("\n"))
+                    .unwrap_or(&path);
+                cflags.push(format!("--sysroot={}", trimmed_path));
+            }
+            _ => {}
         };
 
         BuildConfig {
--- a/mbedtls-sys/build/build.rs
+++ b/mbedtls-sys/build/build.rs
@@ -103,22 +103,6 @@ impl BuildConfig {
             cflags.push("-fno-stack-protector".into());
         }
 
-        // Determine the sysroot for this compiler so that bindgen
-        // uses the correct headers
-        let compiler = cc::Build::new().get_compiler();
-        let output = compiler.to_command().args(&["--print-sysroot"]).output();
-        match (output, compiler.is_like_gnu()) {
-            (Ok(sysroot), true) => {
-                let path = std::str::from_utf8(&sysroot.stdout).expect("Malformed sysroot");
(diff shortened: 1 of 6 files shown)


landhb:
Side note; building for aarch64-unknown-linux-musl will currently fail do this MbedTLS 2.26.0 issue: https://github.com/ARMmbed/mbedtls/pull/4237.

As a temporary workaround until it's fixed, it's possible to build by setting `MBEDTLS_PLATFORM_SNPRINTF_ALT` in the mbedtls config.

mbedtls-sys/build/build.rs
```diff
@@ -103,6 +103,22 @@ impl BuildConfig {
...
+        let compiler = cc::Build::new().get_compiler();
+        let output = compiler.to_command().args(&["--print-sysroot"]).output();
+        match (output, compiler.is_like_gnu()) {
+            (Ok(sysroot), true) => {
+                let path = std::str::from_utf8(&sysroot.stdout).expect("Malformed sysroot");
+                let trimmed_path = path
+                    .strip_suffix("\r\n")
+                    .or(path.strip_suffix("\n"))
+                    .unwrap_or(&path);
+                cflags.push(format!("--sysroot={}", trimmed_path));
```
jethrogb, line 117: This sends the flag to both bindgen and CMake, I think that's not desirable?

mbedtls-sys/build/build.rs
```diff
@@ -103,6 +103,22 @@ impl BuildConfig {
             cflags.push("-fno-stack-protector".into());
         }
 
+        // Determine the sysroot for this compiler so that bindgen
+        // uses the correct headers
+        let compiler = cc::Build::new().get_compiler();
+        let output = compiler.to_command().args(&["--print-sysroot"]).output();
```
jethrogb, line 109: Please run this only if `compiler.is_like_gnu()`

jethrogb (commented):
Thanks! Can you also add a target that uses this to CI?

mbedtls-sys/build/build.rs
landhb: Edited in ce2864a.

mbedtls-sys/build/build.rs
landhb: Edited in ce2864a.

landhb:
> Thanks! Can you also add a target that uses this to CI?

I'll see if I can add the `aarch64-unknown-linux-musl` target to the CI.

landhb:
[Latest build on Travis](https://www.travis-ci.com/github/landhb/rust-mbedtls/builds/222685972)

The three Aarch64 tests are slower since it's running in Qemu.

jethrogb:
I pushed some changes and now some builds are failing and it's entirely unclear to me why. ¯\_(ツ)_/¯

landhb:
> I pushed some changes and now some builds are failing and it's entirely unclear to me why. ¯\_(ツ)_/¯

Dang haha, it looks like on the latest nightly the clang-sys crate is failing to build. [Issue was created here an hour ago](https://github.com/KyleMayes/clang-sys/issues/127).

The aarch64 stable build is failing [here](https://github.com/fortanix/rust-mbedtls/blob/master/mbedtls-sys/vendor/include/mbedtls/ssl.h#L64) But I'm not sure what's causing MBEDTLS_ZLIB_SUPPORT to be defined. Since it looks to be undefined in config.rs.

jethrogb:
Man, I was thinking of making that change just because I thought it would be nicer!

I'll merge this after we get a new nightly

landhb:
> Man, I was thinking of making that change just because I thought it would be nicer!
> 
> I'll merge this after we get a new nightly

Sounds good, thanks!

landhb:
> Except, it didn't fix ARM

Yep, looked the the unquoted "-n" was always [evaluating to true](https://travis-ci.com/github/fortanix/rust-mbedtls/jobs/497949470#L1509).

jethrogb:
Do you need this released soon? Otherwise this will release with the next MbedTLS version.

landhb:
> Do you need this released soon? Otherwise this will release with the next MbedTLS version.

Next release sounds good! For now I can just build from the master branch.

Thanks for the help @jethrogb

PR #163, opened 2021-09-16
Acruceru/mbedtls8 async support

AdrianCX:
This is broken into 2 commits, please review per commit.

First one is the preparation needed to port the async support (plus a migration mode feature to allow both mbedtls7 and 8 to coexists)
Second is the actual async support ported from Mohsen's branch (with some changes due to reference counting)

Let me know if this should be merged (either first/both commits or none), I can also keep it as a separate branch for now similar to the 0.7 one.

PR #171, opened 2021-12-09
Context<S>, generic over IO

MabezDev:
This is a MVP implementation of my proposal in https://github.com/fortanix/rust-mbedtls/issues/3#issuecomment-983703952.

* Update wrapper macros to support generic arg(s), not just lifetimes
* Make Context generic over the underlying IO

This does seem to work with my (not yet published) PR in `rust-native-tls`, and I can run some examples.

## Notes

* I think the `Context<S>` implementation is sound, but `HandshakeContext` I am a lot more unsure of. Particularly the `UnsafeFrom` cast in the Sni Callback.
  * I think it works right now because `HandshakeContext` doesn't use the io field, but its definitely not correct.
  * Making  `HandshakeContext` generic is probably the correct way to do it, but doing that also adds a lot of complications with the callbacks.
* I think the wrapper macros can be refactored to not duplicate everything for generics (I just chose the easiest option to get something working.)

I welcome any feedback :)

### DIFF
--- a/mbedtls-sys/vendor/library/ssl_tls.c
+++ b/mbedtls-sys/vendor/library/ssl_tls.c
@@ -3330,6 +3330,7 @@ static void ssl_calc_finished_tls_sha384(
      * However, to avoid stringop-overflow warning in gcc, we have to cast
      * mbedtls_sha512_finish_ret().
      */
+    // this still emits the warning on clang...
     finish_sha384_t finish = (finish_sha384_t)mbedtls_sha512_finish_ret;
     finish( &sha512, padbuf );
 
--- a/mbedtls/examples/client.rs
+++ b/mbedtls/examples/client.rs
@@ -33,7 +33,9 @@ fn result_main(addr: &str) -> TlsResult<()> {
     config.set_ca_list(cert, None);
     let mut ctx = Context::new(Arc::new(config));
 
+    println!("connecting..");
     let conn = TcpStream::connect(addr).unwrap();
+    println!("establishing SSL connection...");
     ctx.establish(conn, None)?;
 
     let mut line = String::new();
--- a/mbedtls/src/ssl/config.rs
+++ b/mbedtls/src/ssl/config.rs
@@ -320,6 +320,7 @@ impl Config {
             // - We can pointer cast to it to allow storing additional objects.
             //
             let cb = &mut *(closure as *mut F);
+            // TODO is this cast still safe?
             let context = UnsafeFrom::from(ctx).unwrap();
             
             let mut ctx = HandshakeContext::init(context);
--- a/mbedtls/src/ssl/context.rs
+++ b/mbedtls/src/ssl/context.rs
@@ -70,12 +70,12 @@ impl<IO: Read + Write + 'static> IoCallback for IO {
 define!(
     #[c_ty(ssl_context)]
     #[repr(C)]
-    struct Context {
+    struct Context<S> {
         // config is used read-only for mutliple contexts and is immutable once configured.
         config: Arc<Config>, 
 
         // Must be held in heap and pointer to it as pointer is sent to MbedSSL and can't be re-allocated.
-        io: Option<Box<dyn Any>>,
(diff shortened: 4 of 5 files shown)


jethrogb:
Why is this needed? You need a non-static stream?

MabezDev:
More detail here: https://github.com/fortanix/rust-mbedtls/issues/3#issuecomment-983703952 but the general gist is that `establish` has stricter trait bounds than `rust-native-tls`. These stricter trait bounds are currently required because the stream is stored as `dyn Any`. This PR moves towards making `Context` generic over the stream, thus making the trait bounds for the `Context` dependent on the stream. i.e Context will be `Send` if the underlying stream is also `Send`.

AdrianCX:
Having Context generic over IO is also in review: https://github.com/fortanix/rust-mbedtls/pull/163
That also addresses io_mut() returning Any - it will now return Option reference on underlying template.

That review also provides async support ported from:https://github.com/fortanix/rust-mbedtls/pull/127


@MabezDev let me know if first PR above covers what you need and if you spot anything.

AdrianCX:
> thus making the trait bounds for the Context dependent on the stream. i.e Context will be Send if the underlying stream is also Send.

Are you sure you want this?, you will need send for hyper support for example, async also benefits.

MabezDev:
Alternatively, I can port across the relevant changes and make a new PR (also take a stab and refactoring the `define!` macro)

AdrianCX:
Sorry for the ninja edit.

>  Yes, we definitely want the stream to be Send without it, its not very useful at all, sorry if that was not clear. However because of the more relaxed trait bounds on rust-native-tls's functions, it would be better to have Send automatically derived by the compiler.

Do you have an example where changes are useful for / what breaks on your side if io to be send?)

MabezDev:
> Do you have an example where changes are useful for / what breaks on your side if io is required to be send as it is right now?

The main reason I want to remove the explicit bounds, is because [rust-native-tls does not have them](https://github.com/sfackler/rust-native-tls/blob/5a695e2392e8899fd1f2bf34fce6b84707e9bc92/src/lib.rs#L500), only `Read + Write` bounds. There would be more friction in upstreaming if these bounds had to stay explicit.

> They also help in general with integration with other libraries like hyper.

Could you explain how this helps? If hyper requires explicit trait bounds for `Send` that's fine, but with the updates to `Context` provided the stream is `Send`, `Send` will be auto-derived by the compiler and that trait bound will be satisfied automatically (Note: I just pushed a commit where I had forgetten to put the `where S: Send` bounds on the `Context`'s `Send` impl). If the stream is _not_ send, instead of the compiler error being in rust-mbedtls, it will now be in hyper (and the compiler should give a break down of what field causes the `!Send`ness, all the way back to the stream.)

AdrianCX:
Makes sense, I'll look into adding that into my PR and just require that object is static.

MabezDev:
Closing in favour of #163

AdrianCX:
@MabezDev I assume this is what you wanted? https://github.com/fortanix/rust-mbedtls/pull/186

AdrianCX:
@MabezDev please also update here once you have the rust-native-tls PR raised or crate available somewhere. I was looking at doing that but will put that on hold for now since you've already started on it.

AdrianCX:
@MabezDev started doing the native-tls work as well, some very WIP code if you're curious:

https://github.com/AdrianCX/rust-native-tls/tree/acruceru/rust-mbedtls

This pretty much plugs in place with tokio-native-tls (need to update some error messages for though)

MabezDev:
Sorry for the delay in getting back to you. My attempt was based on the original PR: https://github.com/sfackler/rust-native-tls/pull/142, just a straight rewrite using `Arc` etc instead of the boxed pointers in that PR. 

Whilst my PR works, not all tests pass and I'm running low on cycles to work on this. I'll get to finishing this eventually, but perhaps you'd like to take the lead from here on out?

Happy to review the PR whenever it's ready, just ping me :)

PR #172, opened 2022-01-05
Adding `Certificate::verify_with_callback` function

raoulstrackx:
For [Nitro attestation verification](https://github.com/fortanix/rust-sgx/pull/376/) we need some features introduced by @AdrianCX on the `acruceru/port-async-support` branch. For the rust-sgx repo, we want to avoid relying on a separate branch for rust-mbedtls. With this PR two features are added to master:
* serialization and deserialization of signatures
* Certificate verification with a callback function

### DIFF
--- a/mbedtls/src/x509/certificate.rs
+++ b/mbedtls/src/x509/certificate.rs
@@ -225,12 +225,12 @@ impl Certificate {
         chain: &MbedtlsList<Certificate>,
         trust_ca: &MbedtlsList<Certificate>,
         err_info: Option<&mut String>,
-        mut cb: Option<F>,
+        cb: Option<F>,
     ) -> Result<()>
     where
         F: VerifyCallback + 'static,
     {
-        let (f_vrfy, p_vrfy): (Option<unsafe extern "C" fn(_, _, _, _) -> _>, _) = if let Some(cb) = cb.as_mut() {
+        let (f_vrfy, p_vrfy): (Option<unsafe extern "C" fn(_, _, _, _) -> _>, _) = if let Some(cb) = cb.as_ref() {
             (Some(x509::verify_callback::<F>),
             cb as *const _ as *mut c_void)
         } else {
--- a/mbedtls/src/x509/certificate.rs
+++ b/mbedtls/src/x509/certificate.rs
@@ -11,7 +11,7 @@ use core::iter::FromIterator;
 use core::ptr::NonNull;
 
 use mbedtls_sys::*;
-use mbedtls_sys::types::raw_types::*;
+use mbedtls_sys::types::raw_types::{c_char, c_void};
 
 use crate::alloc::{List as MbedtlsList, Box as MbedtlsBox};
 #[cfg(not(feature = "std"))]
@@ -21,8 +21,7 @@ use crate::hash::Type as MdType;
 use crate::pk::Pk;
 use crate::private::UnsafeFrom;
 use crate::rng::Random;
-use crate::x509::Time;
-use crate::x509::VerifyError;
+use crate::x509::{self, Time, VerifyCallback};
 
 extern "C" {
     pub(crate) fn forward_mbedtls_calloc(n: mbedtls_sys::types::size_t, size: mbedtls_sys::types::size_t) -> *mut mbedtls_sys::types::raw_types::c_void;
@@ -222,11 +221,21 @@ impl Certificate {
         MdType::from(self.inner.sig_md)
     }
 
-    pub fn verify(
+    fn verify_ex<F>(
         chain: &MbedtlsList<Certificate>,
         trust_ca: &MbedtlsList<Certificate>,
         err_info: Option<&mut String>,
-    ) -> Result<()> {
+        cb: Option<F>,
+    ) -> Result<()>
+    where
+        F: VerifyCallback + 'static,
+    {
+        let (f_vrfy, p_vrfy): (Option<unsafe extern "C" fn(_, _, _, _) -> _>, _) = if let Some(cb) = cb {
+            (Some(x509::verify_callback::<F>),
+            &cb as *const _ as *mut c_void)
+        } else {
+            (None, ::core::ptr::null_mut())
+        };
         let mut flags = 0;
         let result = unsafe {
             x509_crt_verify(
@@ -235,8 +244,8 @@ impl Certificate {
                 ::core::ptr::null_mut(),
                 ::core::ptr::null(),
                 &mut flags,
-                None,
-                ::core::ptr::null_mut(),
+                f_vrfy,
+                p_vrfy,
             )
         }
         .into_result();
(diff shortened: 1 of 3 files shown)


mbedtls/src/pk/dsa/mod.rs
```diff
@@ -218,8 +218,12 @@ fn sample_secret_value<F: Random>(upper_bound: &Mpi, rng: &mut F) -> Result<Mpi>
 }
 
 fn encode_dsa_signature(r: &Mpi, s: &Mpi) -> Result<Vec<u8>> {
-    let r = BigUint::from_bytes_be(&r.to_binary()?);
-    let s = BigUint::from_bytes_be(&s.to_binary()?);
+    serialize_signature(&r.to_binary()?, &s.to_binary()?)
+}
+
+pub fn serialize_signature(r: &[u8], s: &[u8]) -> Result<Vec<u8>> {
```
Pagten, line 224: This function only outputs a correct DSA signature if `r` and `s` are the _r_ and _s_ components of a DSA signature (that part is obvious from the context), encoded as specified in [https://datatracker.ietf.org/doc/html/rfc8017#section-4.1](https://datatracker.ietf.org/doc/html/rfc8017#section-4.1). I think it would be good to add that as rust documentation on this function and perhaps a similar rustdoc on `deserialize_signature()`.

mbedtls/src/x509/certificate.rs
```diff
@@ -253,6 +254,44 @@ impl Certificate {
         }
         result.map(|_| ())
     }
+
+    pub fn verify_callback<F>(
```
Pagten, line 258: Not sure if we're following some pre-existing function naming scheme, but otherwise `verify_with_callback()` seems a more intuitive name to me.

mbedtls/src/x509/certificate.rs
```diff
@@ -253,6 +254,44 @@ impl Certificate {
...
+                    let prefix = "\0";
+                    x509_crt_verify_info(buf, size, prefix.as_ptr() as *const _, flags)
+                });
+                if let Ok(error_str) = verify_info {
+                    *err_info = error_str;
+                }
+            }
+        }
+        result.map(|_| ())
+    }
```
Pagten, line 293: Could we avoid the code duplication between the existing `fn verify()` and this new `fn verify_callback()` by having both these functions call a private helper function that takes an `Option<F>` as argument?

mbedtls/src/x509/certificate.rs
```diff
@@ -719,6 +758,38 @@ impl Extend<MbedtlsBox::<Certificate>> for MbedtlsList<Certificate> {
     }
 }
 
+pub(crate) unsafe extern "C" fn verify_callback<F>(
```
Pagten, line 761: Could we reuse this code in `ssl::config::set_verify_callback()`?

mbedtls/src/x509/certificate.rs
```diff
@@ -1004,7 +1089,18 @@ cYp0bH/RcPTC0Z+ZaqSWMtfxRrk63MJQF9EXpDCdvQRcTMD9D85DJrMKn8aumq0M
...
+                Ok(())
+            };
+
+            let mut err_str = String::new();
+            let res = Certificate::verify_callback(&chain, &mut c_root, Some(&mut err_str), verify_callback);
+
+            match res {
+                Ok(()) => (),
+                Err(e) => assert!(false, "Failed to verify, error: {}, err_str: {}", e, err_str),
+            };
```
Pagten, line 1103: Shouldn't we keep the original test (which uses `Certificate::verify()` instead of `Certificate::verify_callback()`) as well?

Pagten (changes requested):

mbedtls/src/x509/certificate.rs
raoulstrackx: I agree that it's a better function name and changed it for now. I'm also not that familiar with the code base. Somebody with more experience may want to pitch in.

mbedtls/src/pk/dsa/mod.rs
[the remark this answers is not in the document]
raoulstrackx: Removed these functions, will open another PR for `pkix`

mbedtls/src/x509/certificate.rs
```diff
@@ -221,11 +221,21 @@ impl Certificate {
...
         chain: &MbedtlsList<Certificate>,
         trust_ca: &MbedtlsList<Certificate>,
         err_info: Option<&mut String>,
-    ) -> Result<()> {
+        cb: Option<F>,
+    ) -> Result<()>
+    where
+        F: VerifyCallback + 'static,
+    {
+        let (f_vrfy, p_vrfy): (Option<unsafe extern "C" fn(_, _, _, _) -> _>, _) = if let Some(cb) = cb {
```
AdrianCX, line 233: Can cb die at line 236?, since this is take by value?, which would mean the pointer is invalid?

maybe:
```
if let Some(cb) = cb.as_ref()
```

and adjust line 235?

mbedtls/src/x509/certificate.rs
```diff
@@ -253,6 +263,27 @@ impl Certificate {
...
         result.map(|_| ())
     }
+
+    pub fn verify(
+        chain: &MbedtlsList<Certificate>,
+        trust_ca: &MbedtlsList<Certificate>,
+        err_info: Option<&mut String>,
+    ) -> Result<()> {
+        let none: Option<Box<dyn VerifyCallback + 'static>> = None;
+        Self::verify_ex(chain, trust_ca, err_info, none)
```
jethrogb, line 273:
```suggestion
        Self::verify_ex(chain, trust_ca, err_info, None::<&dyn VerifyCallback>)
```

mbedtls/src/x509/mod.rs
```diff
@@ -117,6 +119,39 @@ impl VerifyError {
...
+    flags: *mut u32,
+) -> c_int
+where
+    F: VerifyCallback + 'static,
+{
+    if crt.is_null() || closure.is_null() || flags.is_null() {
+        return ::mbedtls_sys::ERR_X509_BAD_INPUT_DATA;
+    }
+
+    let cb = &mut *(closure as *mut F);
```
jethrogb, line 137: This has type `&mut F` but you passed in `&F` above in `fn verify_ex`

mbedtls/src/x509/certificate.rs
raoulstrackx: Oh nice! I didn't know about that syntax

mbedtls/src/x509/certificate.rs
```diff
@@ -221,11 +221,21 @@ impl Certificate {
...
         chain: &MbedtlsList<Certificate>,
         trust_ca: &MbedtlsList<Certificate>,
         err_info: Option<&mut String>,
-    ) -> Result<()> {
+        cb: Option<&F>,
+    ) -> Result<()>
+    where
+        F: VerifyCallback + 'static,
+    {
+        let (f_vrfy, p_vrfy): (Option<unsafe extern "C" fn(_, _, _, _) -> _>, _) = if let Some(cb) = cb.as_ref() {
```
AdrianCX, line 233: if you use
```
cb: Option<&F>,
```

Then you shouldn't:
```
if let Some(cb) = cb.as_ref()
```

cb would now be a Option<&&F>

if you then cast this cb via:
```
cb as *const _ as *mut c_void)
```

Do you get a pointer to a pointer, or a proper pointer?

This would be solved either if you pass cb as Option<F> or not use as_ref here?

PR #173, opened 2022-01-05
Update to mbedtls 2.28.0

MihirLuthra:


mbedtls-sys/build/config.rs
```diff
@@ -146,6 +146,7 @@ const DEFAULT_DEFINES: &'static [CDefine] = &[
     ("MBEDTLS_AES_ROM_TABLES",                            Undefined),
     ("MBEDTLS_AES_FEWER_TABLES",                          Undefined),
     ("MBEDTLS_CAMELLIA_SMALL_MEMORY",                     Undefined),
+    ("MBEDTLS_CHECK_RETURN_WARNING",                      Undefined),
```
jethrogb, line 149: Let's set this. Bindgen ignores it but it may be able to use it in the futures.

mbedtls-sys/build/config.rs
```diff
@@ -377,6 +379,7 @@ const DEFAULT_DEFINES: &'static [CDefine] = &[
     ("MBEDTLS_PLATFORM_VSNPRINTF_MACRO",                  Undefined), // default: vsnprintf
     ("MBEDTLS_PLATFORM_NV_SEED_READ_MACRO",               Undefined), // default: mbedtls_platform_std_nv_seed_read
     ("MBEDTLS_PLATFORM_NV_SEED_WRITE_MACRO",              Undefined), // default: mbedtls_platform_std_nv_seed_write
+    ("MBEDTLS_CHECK_RETURN",                              Undefined), // default: __attribute__((__warn_unused_result__))
```
jethrogb, line 382: Missing MBEDTLS_IGNORE_RETURN

mbedtls/src/pk/mod.rs
```diff
@@ -1259,23 +1259,42 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
         ];
 
         for digest in &digests {
+            let digest = *digest;
```
jethrogb: Making changes to the test is not appropriate as this implies a public API change, which is backwards-incompatible. Instead, change `Pk::sign` to deal with the new behavior. The release notes imply this is a RSA-specific change, so please also add some ECDSA tests to make sure you're not making any changes to the behavior there.

jethrogb: Also please add a test with `hash=&[]` as the release notes indicate special before if zero.

MihirLuthra: What kind of changes are you expecting in `Pk::sign`? AFAIU, the change was a bugfix for a behaviour that was already incorrect. This change was backported in `2.26` requiring the argument `hash_len` to be at least the length of message digest. In `2.28`, the length needs to be equal.
Passing a lesser length is surely a bug but we could probably add a patch to `vendor` to accept `hash_len` which is at least the length of message digest instead of equal and bump the patch version of `mbedtls` crate?

References:
[1] https://github.com/ARMmbed/mbedtls/pull/4561
[2] https://github.com/ARMmbed/mbedtls/pull/4707

jethrogb: When I wrote my comment I was thinking we could make `hash` shorter in sign/verify. But actually, silently signing less than the full hash input by the user is also bad. So I think we need to make the incompatible change for security reasons. I've simplified your changes though (force-pushed onto your branch), please review.

MihirLuthra: Thanks, those changes LGTM.

mbedtls-sys/build/config.rs
MihirLuthra: As discussed, not adding this as this is a function-like macro. For reference:
```
/** \def MBEDTLS_IGNORE_RETURN
 *
 * This macro requires one argument, which should be a C function call.
 * If that function call would cause a #MBEDTLS_CHECK_RETURN warning, this
 * warning is suppressed.
 */
//#define MBEDTLS_IGNORE_RETURN( result ) ((void) !(result))
```

PR #175, opened 2022-01-11
Use SHA-512 for aarch64-linux-musl-cross check

raoulstrackx:
Replacing SHA1 for verifying downloaded `aarch64-linux-musl-cross` compiler


aarch64-linux-musl-cross.tgz.sha512sum
```diff
@@ -0,0 +1 @@
+c8ee0e7fd58f5ec6811e3cec5fcdd8fc47cb2b49fb50e9d7717696ddb69c812547b5f389558f62dfbf9db7d6ad808a5a515cc466b8ea3e9ab3daeb20ba1adf33  /tmp/aarch64-linux-musl-cross.tgz
```
jethrogb, line 1: Actually this is not a good location for this file. Just do `echo ... | sha512sum -c` instead

raoulstrackx: Yes that looks nicer!

PR #177, opened 2022-01-15
Create in-place variants for encrypt_auth/decrypt_auth

kedars:
This is invaluable for embedded systems

### DIFF
--- a/mbedtls/src/cipher/mod.rs
+++ b/mbedtls/src/cipher/mod.rs
@@ -286,6 +286,20 @@ impl Cipher<Encryption, Authenticated, AdditionalData> {
             self.change_state(),
         ))
     }
+
+    pub fn encrypt_auth_inplace(
+        mut self,
+        ad: &[u8],
+        data_and_tag: &mut [u8],
+        input_len: usize,
+        tag_len: usize,
+    ) -> Result<(usize, Cipher<Encryption, Authenticated, Finished>)> {
+        Ok((
+            self.raw_cipher
+                .encrypt_auth_inplace(ad, data_and_tag, input_len, tag_len)?,
+            self.change_state(),
+        ))
+    }
 }
 
 impl Cipher<Decryption, Authenticated, AdditionalData> {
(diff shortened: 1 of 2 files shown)


kedars:
ping... any review comments for including this?

jethrogb:
I think it's a bit weird to have the in-place version for the single-step authenticated crypto functions because input and output sizes don't match. I'd be supportive of an in-place version of `update`.

kedars:
How about splitting the tag into a separate out parameter for the in-place variant? The plain/cipher sizes would then match, and the caller gets the control

jethrogb:
I'm not sure the MbedTLS API supports that?

kedars:
Took a stab at it, let me know what you think

kedars:
For the `update` operation, `mbedtls_cipher_update()`, the [mbedTLS documentation](https://tls.mbed.org/api/cipher_8h.html#a959598dfef9d7f2add8861e0c518f303) specifically forbids the output and input buffers to be the same.

mbedtls/src/cipher/mod.rs
```diff
@@ -401,6 +427,44 @@ fn ccm() {
...
+    let cipher = Cipher::<_, Authenticated, _>::new(
+        raw::CipherId::Aes,
+        raw::CipherMode::CCM,
+        (k.len() * 8) as _,
+    )
+    .unwrap();
+    let cipher = cipher.set_key_iv(&k, &iv).unwrap();
+    let (data, tag) = c.split_at_mut(4);
+    cipher.decrypt_auth_inplace(&ad, data, tag).unwrap();
+    assert_eq!(validate_plain, c[0..4]);
```
jethrogb, line 465:
```suggestion
    assert_eq!(validate_plain, data);
```

jethrogb (commented):
Looks good now! One small nit.

mbedtls/src/cipher/mod.rs
kedars: makes sense, done 👍

PR #182, opened 2022-02-08
Initial changes for async support

AdrianCX:
Splitting https://github.com/fortanix/rust-mbedtls/pull/163 into separate parts.

Also rebased since @raoulstrackx already helped with some pieces of the PR above.


jethrogb:
> verify with sig len 0 makes mbedtls assume default siglen and potentially overflow instead of stopping.

This is fixed in MbedTLS 2.28, I think?

mbedtls/src/pk/mod.rs
```diff
@@ -913,6 +886,10 @@ impl Pk {
     }
 
     pub fn verify(&mut self, md: MdType, hash: &[u8], sig: &[u8]) -> Result<()> {
+        if hash.len() == 0 || sig.len() == 0 {
```
jethrogb, line 889: For hash, please add a test. For sig, is this necessary?

mbedtls/src/ssl/config.rs
```diff
@@ -100,6 +100,49 @@ callback!(DbgCallback: Fn(i32, Cow<'_, str>, i32, Cow<'_, str>) -> ());
...
+
+unsafe impl Send for NullTerminatedStrList {}
+unsafe impl Sync for NullTerminatedStrList {}
+
+impl NullTerminatedStrList {
+    pub fn new(list: &[&str]) -> Result<Self> {
+        let mut c = Vec::with_capacity(list.len() + 1);
+
+        for s in list {
+            let cstr = ::std::ffi::CString::new(*s).map_err(|_| Error::SslBadInputData)?;
```
jethrogb, line 117: Will cause memory leak if anything after the 0th element errors. I think it's ok to have `NullTerminatedStrList` contain a Vec instead of a boxed slice. Then you can rely on the `NullTerminatedStrList` destructor while filling.

mbedtls/src/ssl/config.rs
```diff
@@ -100,6 +100,49 @@ callback!(DbgCallback: Fn(i32, Cow<'_, str>, i32, Cow<'_, str>) -> ());
...
+#[repr(transparent)]
+pub struct NullTerminatedStrList {
+    c: Box<[*mut i8]>,
+}
+
+unsafe impl Send for NullTerminatedStrList {}
+unsafe impl Sync for NullTerminatedStrList {}
+
+impl NullTerminatedStrList {
+    pub fn new(list: &[&str]) -> Result<Self> {
```
jethrogb, line 113: I think this needs the `std` feature due to the use of `CString`?

mbedtls/src/ssl/config.rs
```diff
@@ -100,6 +100,49 @@ callback!(DbgCallback: Fn(i32, Cow<'_, str>, i32, Cow<'_, str>) -> ());
 callback!(SniCallback: Fn(&mut HandshakeContext, &[u8]) -> Result<()>);
 callback!(CaCallback: Fn(&MbedtlsList<Certificate>) -> Result<MbedtlsList<Certificate>>);
 
+
+#[repr(transparent)]
+pub struct NullTerminatedStrList {
+    c: Box<[*mut i8]>,
```
jethrogb, line 106: Should be `c_char`, not `i8`.

mbedtls/src/ssl/config.rs
```diff
@@ -184,6 +225,20 @@ impl Config {
         self.ciphersuites.push(list);
     }
 
+    /// Set the supported Application Layer Protocols.
+    ///
+    /// Each protocol name in the list must also be terminated with a null character (`\0`).
```
jethrogb, line 230: Null terminator is not required as we add it.

mbedtls/src/ssl/context.rs
```diff
@@ -99,19 +110,22 @@ impl Context {
...
-            handshake_crl: None,
-            
-            handshake_cert: vec![],
-            handshake_pk: vec![],
         }
     }
+}
 
-    pub fn establish<T: IoCallback + Send + Sync + 'static>(&mut self, io: T, hostname: Option<&str>) -> Result<()> {
+impl<T: IoCallback + Send + Sync + 'static> Context<T> {
```
jethrogb, line 127: I thought part of the point was to get rid of the `Send + Sync + 'static` requirement here?

mbedtls/src/ssl/context.rs
```diff
@@ -127,21 +141,35 @@ impl Context {
...
-                Err(e) => {
-                    // safely end borrow of io
-                    ssl_set_bio(self.into(), ::core::ptr::null_mut(), None, None, None);
-                    self.io = None;
-                    Err(e)
-                },
-                Ok(_) => {
-                    Ok(())
+impl<T> Context<T> {
+    pub fn handshake(&mut self) -> Result<()> {
```
jethrogb, line 152: This shouldn't be `pub`, I think?

mbedtls/src/ssl/context.rs
```diff
@@ -70,13 +71,7 @@ impl<IO: Read + Write + 'static> IoCallback for IO {
...
     #[c_ty(ssl_context)]
     #[repr(C)]
-    struct Context {
-        // config is used read-only for mutliple contexts and is immutable once configured.
-        config: Arc<Config>, 
-
-        // Must be held in heap and pointer to it as pointer is sent to MbedSSL and can't be re-allocated.
-        io: Option<Box<dyn Any>>,
-        
+    struct HandshakeContext {
```
jethrogb, line 74: Please add a test that this type is !Sync. I think we don't expose owned values, so checking !Send isn't necessary.

mbedtls/src/ssl/context.rs
```diff
@@ -187,22 +215,21 @@ impl Context {
             self.io = None;
         }
     }
-
```
jethrogb, line 190: Don't remove newline

mbedtls/src/ssl/context.rs
```diff
@@ -260,7 +301,7 @@ impl Drop for Context {
     }
 }
 
-impl Read for Context {
+impl<T: Read> Read for Context<T> {
```
jethrogb, line 304: Bit of a weird type contraint since we never call read. Either don't add a constraint, or add a constraint on IoCallback. Same for Write.

mbedtls/src/ssl/context.rs
```diff
@@ -301,42 +336,44 @@ pub struct HandshakeContext<'ctx> {
 // - mbedtls not providing any callbacks on handshake finish.
 // - no reasonable way to obtain a storage within the sni callback tied to the handshake or to the rust Context. (without resorting to a unscalable map or pointer magic that mbedtls may invalidate)
 //
-impl<'ctx> HandshakeContext<'ctx> {
-
-    pub(crate) fn init(context: &'ctx mut Context) -> Self {
-        HandshakeContext { context }
+impl HandshakeContext {
+    pub fn reset_handshake(&mut self) {
```
jethrogb, line 340: I don't think this should be pub.

mbedtls/src/ssl/context.rs
```diff
@@ -301,42 +336,44 @@ pub struct HandshakeContext<'ctx> {
...
         
-        unsafe { ssl_set_hs_authmode(self.context.into(), am as i32) }
+        unsafe { ssl_set_hs_authmode(self.into(), am as i32) }
         Ok(())
     }
 
     pub fn set_ca_list(
         &mut self,
-        chain: Arc<MbedtlsList<Certificate>>,
+        chain: Option<Arc<MbedtlsList<Certificate>>>,
```
jethrogb, line 358: This change doesn't make sense to me. Is the intention to set only the CRL? The behavior when passing None should be clearly documented.

mbedtls/src/wrapper_macros.rs
```diff
@@ -109,6 +113,32 @@ macro_rules! define_enum {
...
+        pub struct $name<$($l)*> {
+            $($(#[$mm])* $member: $member_type,)*
+        }
+        );
+
+        as_item!(
+        #[allow(dead_code)]
+        impl<$($l)*> $name<$($l)*> {
+            pub(crate) fn handle(&self) -> &::mbedtls_sys::$inner {
+                self.inner.handle()
```
jethrogb, line 129: It's not appropriate to use `inner` here since that field is defined outside of the macro.

I don't think this macro adds much for the case where we're wrapping another type that's already using the macro. Just implement it by hand. No need to `impl Send` I think as it's automatically derived.

mbedtls/Cargo.toml
```diff
@@ -36,7 +36,7 @@ rs-libc = "0.2.0"
 chrono = "0.4"
 
 [dependencies.mbedtls-sys-auto]
-version = "2.25.0"
+version = "2.26.0"
```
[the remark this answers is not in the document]
AdrianCX: I want it to match the current mbedtls-sys-auto as I did run into some cargo issues.

I'll update to the latest version specified there.

(or can I remove the version line?

mbedtls/src/pk/mod.rs
```diff
@@ -201,34 +201,7 @@ define!(
...
-//                    mbedtls_ecp_mul_restartable: ../../../mbedtls-sys/vendor/crypto/library/ecp.c:2351
-//                        MBEDTLS_ECP_INTERNAL_ALT is not defined. (otherwise it might not be safe depending on ecp_init/ecp_free) ../../../mbedtls-sys/build/config.rs:131
-//                        Passes as const to: mbedtls_ecp_check_privkey / mbedtls_ecp_check_pubkey / mbedtls_ecp_get_type( grp
-//        
-// - Ignored due to not defined: ecdsa_verify_rs_wrap, ecdsa_sign_rs_wrap, ecdsa_rs_alloc, ecdsa_rs_free
-//   (Undefined - MBEDTLS_ECP_RESTARTABLE - ../../../mbedtls-sys/build/config.rs:173)
-//
-// - Only const access to context: eckey_check_pair
-//
-// - Only used when creating/freeing - which is safe by design: ecdsa_alloc_wrap, ecdsa_free_wrap
```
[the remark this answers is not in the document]
AdrianCX: Reverting, I thought the new comment was clearer

mbedtls/src/ssl/context.rs
AdrianCX: This is needed for that work and async.

I will raise another PR with that part.

mbedtls/src/ssl/context.rs
AdrianCX: It allows unsetting both the crl and ca chain (setting them to null)

This helps if they were previously set and they need to be altered.

mbedtls/src/wrapper_macros.rs
AdrianCX: Handle is useful to integrate with other common macros from same file such as below:

```
    { into $name:ident inner $inner:ident $(lifetime $l:tt)* lifetime2 $l2:tt } => {
        as_item!(
        impl<$l2,$($l),*> Into<*const $inner> for &$l2 $name<$($l)*> {
            fn into(self) -> *const $inner {
                self.handle()
            }
        }
        );
```

jethrogb: So implement handle manually.

mbedtls/src/pk/mod.rs
```diff
@@ -1274,6 +1289,18 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
                 )
                 .unwrap();
             pk.verify(digest, data, &signature[0..len]).unwrap();
+
+            assert_eq!(pk.verify(digest, data, &[]).unwrap_err(), Error::PkBadInputData);
+            assert_eq!(pk.verify(digest, &[], &signature[0..len]).unwrap_err(), Error::PkBadInputData);
+
+
+            let mut dummy_sig = vec![];
```
jethrogb, line 1297: Nit: doesn't need to be Vec.

mbedtls/src/ssl/config.rs
```diff
@@ -184,6 +220,18 @@ impl Config {
         self.ciphersuites.push(list);
     }
 
+    /// Set the supported Application Layer Protocols.
+    pub fn set_alpn_protocols(&mut self, protocols: Arc<NullTerminatedStrList>) -> Result<()> {
+        unsafe {
+            ssl_conf_alpn_protocols(&mut self.inner, protocols.as_ptr() as *mut _)
```
jethrogb, line 226: Why as mut? It's const in the header I'm looking at.

mbedtls/src/ssl/context.rs
```diff
@@ -89,7 +84,23 @@ define!(
     impl<'a> UnsafeFrom<ptr> {}
 );
 
-impl Context {
+define!(
+    #[c_custom_ty(ssl_context)]
```
jethrogb, line 88: We discussed not using a macro here

mbedtls/src/ssl/context.rs
```diff
@@ -127,21 +149,35 @@ impl Context {
...
-                Err(e) => {
-                    // safely end borrow of io
-                    ssl_set_bio(self.into(), ::core::ptr::null_mut(), None, None, None);
-                    self.io = None;
-                    Err(e)
-                },
-                Ok(_) => {
-                    Ok(())
+impl<T> Context<T> {
+    fn handshake(&mut self) -> Result<()> {
```
jethrogb, line 160: Since we own `T` I think it's not really necessary to unset the IO in case of error anymore. This greatly simplifies the function.

mbedtls/src/ssl/context.rs
```diff
@@ -350,21 +397,31 @@ impl<'ctx> HandshakeContext<'ctx> {
...
+        self.handshake_pk.push(key);
 
         Ok(())
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use crate::ssl::context::HandshakeContext;
+    use crate::tests::{TestTrait}; //, Testable};
```
jethrogb, line 418: Stray comment

PR #183, opened 2022-02-09
Add incremental hmac new/update/finish support

kedars:


mbedtls/src/hash/mod.rs
```diff
@@ -125,6 +125,36 @@ impl Md {
         }
     }
 
+    pub fn new_hmac(md: Type, key: &[u8]) -> Result<Md> {
```
jethrogb, line 128: This should be a separate type. It shouldn't be possible to mix hash/hmac functions. The hmac function you added recently should be on the separate type as well.

kedars:
I had thought of it that way, but the underlying md-context is shared in the mbedTLS APIs. 

The latest code has it done that way, thoughts?

kedars:
Or we could just duplicate `Md::new()` as `Hmac::new()` with the necessary changes and then use 
```
pub struct Hmac {
    ctx: Md,
}
```

jethrogb:
Your implementation of `Hmac` is mostly fine. But it's not necessary to split Md/MdCtx

kedars:
Updated, also put `hkdf()` in its own, so there won't be a namespace conflict when it is moved later on

PR #187, opened 2022-02-17
Enable ARIA cipher

zugzwang:


jethrogb (commented):
Missing ARIA TLS cipher suites in mbedtls/src/ssl/ciphersuites.rs.

mbedtls/src/ssl/ciphersuites.rs
```diff
@@ -121,6 +121,44 @@ define!(
         EcdhePskWithNullSha = TLS_ECDHE_PSK_WITH_NULL_SHA,
         EcdhePskWithNullSha256 = TLS_ECDHE_PSK_WITH_NULL_SHA256,
         EcdhePskWithNullSha384 = TLS_ECDHE_PSK_WITH_NULL_SHA384,
+        RsaWithAria128CbcSha256 = TLS_RSA_WITH_ARIA_128_CBC_SHA256,
```
zugzwang, line 124: To reviewer:
The ciphersuites in this file are kind of ordered by hash function. But then Camellia suites are almost all together, so I simply put all ARIA ciphersuites together too, as displayed in mbedtls-sys/vendor/include/mbedtls/config.h.

PR #188, opened 2022-02-21
fix pkcs12_rc2

MabezDev:
Not sure how this isn't picked up on CI, but locally I run into this error when I enable the `pkcs12_rc2` feature.

```rust
Updating crates.io index
error: no matching package named `block-cipher-trait` found
location searched: registry `crates-io`
required by package `rc2 v0.3.0`
    ... which satisfies dependency `rc2 = "^0.3"` of package `mbedtls v0.8.2
```


Taowyoo:
Close this since this already been fixed in https://github.com/fortanix/rust-mbedtls/issues/228

PR #190, opened 2022-02-23
Some changes needed to interact with rust-native-tls

AdrianCX:
Update PKCS for newer RC2 since the version used was yanked.

Improved APLN NullTerminatedList to take a ref to Vec<String> as well.

Tokio integration requires proper WouldBlock IO Error.

handshake needs to be pub to be called from tokio-native-tls.

from_pem is annoying in that it needs last byte 0, added code to detect improper pems and fix them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -83,21 +83,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a606a02debe2813760609f57a64a2ffd27d9fdf5b2f133eaca0b248dd92cdd2"
 
 [[package]]
-name = "block-cipher-trait"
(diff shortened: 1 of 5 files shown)


Taowyoo:
close this because this PR is outdated.

- Update PKCS for newer RC2 since the version used was yanked. (done)
- Improved APLN NullTerminatedList to take a ref to Vec as well. (api changed)
- Tokio integration requires proper WouldBlock IO Error. (done)
- handshake needs to be pub to be called from tokio-native-tls. (done)
- from_pem is annoying in that it needs last byte 0, added code to detect improper pems and fix them. (please put this in a new PR)

PR #192, opened 2022-03-09
Mpi::is_probably_prime (mbedtls_mpi_is_prime_ext)

zugzwang:
Support mbedtls_mpi_is_prime, the Miller-Rabin primality test.

PR #194, opened 2022-05-04
Fix checks on curve parameters

zugzwang:
On instantiating a curve by parameters, the code checks that all
field elements are non-zero in 𝔽ₚ. This is not necessary, and forbids
useful curves such as e.g. the Barreto-Naehrig pairing-friendly curves,
of the form y² = x³ + 257 (e.g. `a = 0`), and secp256k1 (used in Bitcoin, but can be instantiated by name).

### DIFF
--- a/mbedtls/src/ecp/mod.rs
+++ b/mbedtls/src/ecp/mod.rs
@@ -96,13 +96,13 @@ impl EcGroup {
         let zero = Mpi::new(0)?;
 
         // basic bounds checking
-        if &a <= &zero
+        if &a < &zero
             || &a >= &p
-            || &b <= &zero
+            || &b < &zero
             || &b >= &p
-            || &g_x <= &zero
+            || &g_x < &zero
             || &g_x >= &p
-            || &g_y <= &zero
+            || &g_y < &zero
             || &g_y >= &p
             || &order <= &zero
         {
--- a/mbedtls/src/ecp/mod.rs
+++ b/mbedtls/src/ecp/mod.rs
@@ -191,6 +191,8 @@ impl EcGroup {
         match self.group_id()? {
             EcGroupId::Curve25519 => Ok(8),
             EcGroupId::Curve448 => Ok(4),
+            // Requires a point-counting algorithm such as SEA.
+            EcGroupId::None => Err(Error::EcpFeatureUnavailable),
             _ => Ok(1),
         }
     }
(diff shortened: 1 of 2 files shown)


zugzwang:
Review ping. Thanks!

zugzwang:
This PR also fixes #195

mbedtls/src/ecp/mod.rs
```diff
@@ -421,6 +423,7 @@ mod tests {
     use crate::bignum::Mpi;
     use crate::ecp::{EcGroup, EcPoint};
     use crate::pk::EcGroupId;
+    use super::Result;
```
xinyufort, line 426: This line is causing your recent Travis CI build to fail: https://app.travis-ci.com/github/fortanix/rust-mbedtls/jobs/570370369

xinyufort (commented):
LGTM, but I'm nowhere close to being an expert on elliptic curve cryptography. I heard you're going to get back w Jack Lloyd, so feel free to ping me again if you want an actual approval

aandyl:
I'd suggest adding a disclaimer/warning on `EcGroup::from_parameters` that (1) this functionality is provided as an escape hatch to enable the use of custom curves, but is quite dangerous, and is not provided natively by mbedtls, (2) callers of this function must be ABSOLUTELY CERTAIN the curve parameters are of trusted origin.

I think that disclaimer should have been here all along, and once the disclaimer is taken into consideration, this pull request doesn't meaningfully change the security properties. The new checks are at least slightly permissive -- a = b = 0 shouldn't be allowed and g = (0, 0) shouldn't be allowed. The latter should already be detected by the existing generator check later in the function. Possibly a = b = 0 would be caught somewhere as well. However, while it's nice if we can sanity check some completely invalid cases, we're not going to be able to provide an assurance of the cryptographic strength of the parameters.

zugzwang:
Thanks @xinyufort and @aandyl for the review.

@aandyl, yes, good idea on the disclaimer. let's do that.

Just for completeness, it would be possible to perform a basic check on parameters, but we don't want to do this on every call. If need be, we could create another function, but I'd stick with the disclaimer:
- the curve needs to be non-singular `4a^3 + 27b^2 != 0 (mod p)`, which captures your `a = b = 0` case,
- the order needs to be prime (Miller-Rabin, https://github.com/fortanix/rust-mbedtls/pull/192) and different to `p`,
- against discrete log transfer attacks: request the cofactor too, the order of `p` in the multiplicative `F_{cofactor}*` is at least 20

mbedtls/src/ecp/mod.rs
```diff
@@ -79,6 +79,13 @@ impl EcGroup {
         Ok(ret)
     }
 
+    /// Enables the use of custom curves.
```
jethrogb, line 82: “Initialize an `EcGroup` with custom group parameters.”

PR #107, opened 2020-05-23
Add PSK support

ndokmai:
Add PSK support to the library with one basic test.

PR #205, opened 2022-07-25
Add PSK support

dongcarl:
A rebased and cleaned up version of #107

### DIFF
--- a/mbedtls/src/ssl/config.rs
+++ b/mbedtls/src/ssl/config.rs
@@ -99,6 +99,7 @@ define!(
 callback!(DbgCallback: Fn(i32, Cow<'_, str>, i32, Cow<'_, str>) -> ());
 callback!(SniCallback: Fn(&mut HandshakeContext, &[u8]) -> Result<()>);
 callback!(CaCallback: Fn(&MbedtlsList<Certificate>) -> Result<MbedtlsList<Certificate>>);
+callback!(PskCallback: Fn(&mut HandshakeContext, &str) -> Result<()>);
 
 
 #[repr(transparent)]
@@ -164,6 +165,7 @@ define!(
         sni_callback: Option<Arc<dyn SniCallback + 'static>>,
         ticket_callback: Option<Arc<dyn TicketCallback + 'static>>,
         ca_callback: Option<Arc<dyn CaCallback + 'static>>,
+        psk_callback: Option<Arc<dyn PskCallback + 'static>>,
     };
     const drop: fn(&mut Self) = ssl_config_free;
     impl<'a> Into<ptr> {}
@@ -199,6 +201,7 @@ impl Config {
             sni_callback: None,
             ticket_callback: None,
             ca_callback: None,
+            psk_callback: None,
         }
     }
 
@@ -457,6 +460,42 @@ impl Config {
         self.dbg_callback = Some(Arc::new(cb));
         unsafe { ssl_conf_dbg(self.into(), Some(dbg_callback::<F>), &**self.dbg_callback.as_mut().unwrap() as *const _ as *mut c_void) }
     }
+
+    pub fn set_psk(&mut self, psk: &[u8], psk_identity: &str) -> Result<()> {
+        unsafe { ssl_conf_psk(&mut self.inner,
+                         psk.as_ptr(), psk.len(),
+                         psk_identity.as_ptr(), psk_identity.len())
+            .into_result().map(|_| ())
+        }
+    }
+
+    #[cfg(feature = "std")]
+    pub fn set_psk_callback<F>(&mut self, cb: F)
+        where
+            F: PskCallback + 'static,
+    {
+        unsafe extern "C" fn psk_callback<F>(
+            closure: *mut c_void,
+            ctx: *mut ssl_context,
+            psk_identity: *const c_uchar,
+            identity_len: size_t,
+        ) -> c_int
+        where
+              F: PskCallback + 'static,
+        {
+            let cb = &mut *(closure as *mut F);
+            let ctx = UnsafeFrom::from(ctx).unwrap();
+
+            let psk_identity = std::str::from_utf8_unchecked(from_raw_parts(psk_identity, identity_len));
+            match cb(ctx, psk_identity) {
+                Ok(()) => 0,
+                Err(e) => e.to_int(),
+            }
+        }
+
+        self.psk_callback = Some(Arc::new(cb));
+        unsafe { ssl_conf_psk_cb(self.into(), Some(psk_callback::<F>), &**self.psk_callback.as_mut().unwrap() as *const _ as *mut c_void) }
+    }
 }
 
 // TODO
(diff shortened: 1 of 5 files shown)


dongcarl:
I don't think that the `verify_chain` failure is related, can anyone confirm?

dongcarl:
Ping @raoulstrackx @Pagten

mbedtls/tests/ssl_conf_psk_cb.rs
```diff
@@ -0,0 +1,61 @@
+#![allow(dead_code)]
```
raoulstrackx, line 1: Shouldn't be needed

raoulstrackx:
I also fixed CI in #206, so could you rebase on top of that one?

mbedtls/tests/ssl_conf_psk_cb.rs
```diff
@@ -0,0 +1,61 @@
...
+    let entropy = Arc::new(entropy_new());
+    let rng = Arc::new(CtrDrbg::new(entropy, None)?);
+    let mut config = Config::new(Endpoint::Server, Transport::Stream, Preset::Default);
+    config.set_rng(rng);
+    config.set_psk_callback(psk_callback);
+    let mut ctx = Context::new(Arc::new(config));
+    ctx.establish(conn, None).map(|_| ())
+}
+
+#[cfg(unix)]
```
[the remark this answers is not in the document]
dongcarl: Removed

mbedtls/tests/ssl_conf_psk_cb.rs
dongcarl: Removed

mbedtls/src/ssl/config.rs
```diff
@@ -457,6 +460,41 @@ impl Config {
...
+            psk_identity: *const c_uchar,
+            identity_len: size_t,
+        ) -> c_int
+        where
+              F: PskCallback + 'static,
+        {
+            let cb = &mut *(closure as *mut F);
+            let ctx = UnsafeFrom::from(ctx).unwrap();
+
+            let psk_identity = std::str::from_utf8_unchecked(from_raw_parts(psk_identity, identity_len));
```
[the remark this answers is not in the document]
dongcarl: What's the best thing to do here? `unwrap` a checked `from_utf8` so the error is more apparent?

dongcarl:
Push f95e61d70238c2784cfc9462c258092c7f160021 -> 8162ee403acc40eb5a5e68950d127b04fd9f0a8e
- Addressed
  - https://github.com/fortanix/rust-mbedtls/pull/205#discussion_r936629319
  - https://github.com/fortanix/rust-mbedtls/pull/205#discussion_r936635492

@raoulstrackx let me know about https://github.com/fortanix/rust-mbedtls/pull/205#discussion_r938336268

dongcarl:
Push 8162ee403acc40eb5a5e68950d127b04fd9f0a8e -> b71ec09e04c1e7d769f8e81285bb8d2821818486
- Reverted fixes applied in https://github.com/fortanix/rust-mbedtls/pull/205#issuecomment-1205887372
- Add `#[cfg(feature = "std")]` to `set_psk_callback` since we use `std::str`

After running the CI, we see that we actually do need some kind of attribute guard for this test because `x86_64-fortanix-unknown-sgx` does not have `crate::support::net`, see:
- https://app.travis-ci.com/github/fortanix/rust-mbedtls/jobs/578847394#L556

There are more than one way to do the attribute guard that can be seen in existing tests:
- `mbedtls/tests/ssl_conf_ca_cb.rs` uses `#![allow(dead_code)]` and `#[cfg(unix)]` since `client` and `server` will be dead code on non-unix
- `mbedtls/tests/client_server.rs` uses a module level `#![cfg(not(target_env = "sgx"))]`

I ended up using a module level `#![cfg(not(target_env = "sgx"))]`, but am happy to do either way.

Ping @raoulstrackx

mbedtls/tests/ssl_conf_psk_cb.rs
dongcarl: See update in: https://github.com/fortanix/rust-mbedtls/pull/205#issuecomment-1205973141

mbedtls/tests/ssl_conf_psk_cb.rs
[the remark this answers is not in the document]
dongcarl: See update in: https://github.com/fortanix/rust-mbedtls/pull/205#issuecomment-1205973141

mbedtls/src/ssl/config.rs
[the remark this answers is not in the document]
raoulstrackx: No you could change the type of `PskCallback` to 
```
callback!(PskCallback: Fn(&mut HandshakeContext, &[u8]) -> Result<()>);
```
so there isn't a need to convert it.

mbedtls/tests/support/keys.rs
```diff
@@ -92,3 +92,7 @@ pub const ROOT_CA_KEY: &'static str = concat!(include_str!("./keys/ca.key"),"\0"
 pub const EXPIRED_CERT_SUBJECT: &'static str = "CN=ExpiredNode";
 pub const EXPIRED_CERT: &'static str = concat!(include_str!("./keys/expired.crt"),"\0");
 pub const EXPIRED_KEY: &'static str = concat!(include_str!("./keys/expired.key"),"\0");
+
+pub const PRESHARED_KEY: &'static [u8] = &[
+    234, 206, 151, 23, 219,  21,  71, 144,
+    107,  42,  23, 67, 249, 173, 182, 224 ];
```
raoulstrackx, line 98: Could you pick a preshared key that is not valid utf8 and also contains a zero byte? It may trigger some weird code paths.

raoulstrackx:
> After running the CI, we see that we actually do need some kind of attribute guard for this test because `x86_64-fortanix-unknown-sgx` does not have `crate::support::net`, see:

Oh I thought CI for `x86_64-fortanix-unknown-sgx` was better than it is. It could use some work, but that's out of scope for this PR. A module level is not ideal but I'd say it's acceptable until the CI for sgx gets straightened out.

@zugzwang could you also take a look when these two small comments of mine are addressed?

mbedtls/tests/support/keys.rs
zugzwang: Also, could we add a case with a leading zero byte? I'm certain it is correctly handled, but I have seen way too many leading zero bugs in the wild/

(Orthogonal but also see https://mbed-tls.readthedocs.io/en/latest/security-advisories/advisories/mbedtls-security-advisory-2020-09-3/).

dongcarl: Huh! If it's not valid UTF-8, we should expect an error from `from_utf8`?

What kind of error do we expect for leading zero bytes?

zugzwang: Why is there a requirement that pre-shared keys are valid UTF-8? At least in [RFC4279](https://datatracker.ietf.org/doc/html/rfc4279), only PSK _identities_ are assumed to be UTF-8 (Sec. 5.1). Also Sec. 4 states that PSK are the result of Diffie Hellman computations, so no UTF-8 expected.

zugzwang:
In the meantime we decided to merge #202 since it also added DTLS support.

Taowyoo:
Thank you for contribution.
Because long time no updates on this PR and PSK support is added in https://github.com/fortanix/rust-mbedtls/pull/202.
So I decided to close this PR.

PR #210, opened 2022-09-26
Replace crate `core_io` with `core2` to support more rust toolchain versions

gaojiaqi7:
`core_io` has not been updated for a long time, the latest rust version it supports is 2021-03-25.
To support more rust toolchain versions, we can use `core2` instead which does not depend on specific nightly versions.

Signed-off-by: Jiaqi Gao <jiaqi.gao@intel.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,12 +197,12 @@ dependencies = [
 ]
 
 [[package]]
-name = "core_io"
-version = "0.1.20210325"
+name = "core2"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "97f8932064288cc79feb4d343a399d353a6f6f001e586ece47fe518a9e8507df"
+checksum = "b49ba7ef1ad6107f8824dbe97de947cbaac53c44e7f9756a1fba0d37c1eec505"
 dependencies = [
- "rustc_version",
+ "memchr",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jyao1:
Hello, is there any feedback for this one?

MihirLuthra:
First of all - apologies for the late response on this PR.

I gave a look to `core2` and I don't think we would prefer switching to it. As per its README, `core2` doesn't depend on any specific version because it copies its code from Rust 1.47.0. I don't think silently using code from an older version along with a new toolchain would be a good practice. `core_io` ensures that the code taken from the rust lib corresponds to the toolchain being used. It seems preferable to maintain `core_io` instead of switching to `core2`.

Alternatively, PR #211 proposes to remove `core_io` completely. My only concern there is that `core_io` is more intuitive to use than the replacements made in the PR.

cc: @zugzwang @raoulstrackx @jethrogb

gaojiaqi7:
Thanks @MihirLuthra , I agree with you that using code from an older version along with a new toolchain is not a good choice, but it may take a lot effort to maintain `core_io` to make it work with the latest rust toolchain all the time.
I prefer PR https://github.com/fortanix/rust-mbedtls/pull/211 because it is more suitable for `no_ std` case and simple.

gaojiaqi7:
@MihirLuthra that's great! thank you :)

MihirLuthra:
Closing this in favour of #211.

Issue #208, opened 2022-09-23
Build failed without standard library


gaojiaqi7:
Hi,

I am trying to use mbedtls in `no_std` environment with `x86_64-unknown-none` target without standard library built-in, but I met the following errors when compiling mbedtls:

```
error[E0277]: the trait bound `rust_alloc::boxed::Box<SavedRawCipher>: Deserialize<'_>` is not satisfied
    --> mbedtls/src/cipher/raw/serde.rs:52:9
     |
52   |         inner_cipher_ctx: Box<SavedRawCipher>
     |         ^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `rust_alloc::boxed::Box<SavedRawCipher>`
     |
     = help: the following other types implement trait `Deserialize<'de>`:
               &'a [u8]
               &'a str
               ()
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
               (T0, T1, T2, T3, T4)
               (T0, T1, T2, T3, T4, T5)
             and 102 others
note: required by a bound in `next_value`
    --> /home/tdvf/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.101/src/de/mod.rs:1845:12
     |
1845 |         V: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `next_value`

Some errors have detailed explanations: E0277, E0407.
For more information about an error, try `rustc --explain E0277`.
error: could not compile `mbedtls` due to 4 previous errors
```
```
error[E0405]: cannot find trait `Default` in this scope
    --> /home/tdvf/.cargo/registry/src/github.com-1ecc6299db9ec823/byteorder-1.3.2/src/lib.rs:1831:6
     |
1831 | impl Default for LittleEndian {
     |      ^^^^^^^ not found in this scope
     |
help: consider importing this trait
     |
83   | use core::default::Default;
     |

Some errors have detailed explanations: E0405, E0425, E0463.
For more information about an error, try `rustc --explain E0405`.
error: could not compile `byteorder` due to 173 previous errors
```

The error message is too long, so I only pasted part of it here.
The build command I'm using is:
```
cargo build -p mbedtls -Zbuild-std=core,alloc,compiler_builtins --target=x86_64-unknown-none --no-default-features --features="no_std_deps,rdrand"
```
The same result applies to:
```
[...]

PR #211, opened 2022-10-12
Remove core_io dependency and add alternative interface for no_std environments

DrTobe:
As already stated in https://github.com/fortanix/rust-mbedtls/pull/210, `core_io` "has not been updated for a long time, the latest rust version it supports is 2021-03-25." Since `no_std` usage of `rust-mbedtls` currently depends on `core_io`, this is a major barrier when adopting `rust-mbedtls` in `no_std` environments.

As opposed to https://github.com/fortanix/rust-mbedtls/pull/210, I propose to remove any `std::io` replacement altogether because we need a different interface (non `io::Write`, non `io::Read`) for DTLS anyways. So this interface can be used for TLS in `no_std` environments, too, which removes the necessity to have a `std::io` replacement at all. If you think the benefits are worth the effort, a `std::io` replacement could be added optionally with a feature. I have not had a detailed look at `core2` but I found working with `core_io` very limiting, so maybe `core2` could be a better option.

When working on these changes, I have also fixed some other minor issues with `no_std` usage. Specifically, this should also fix https://github.com/fortanix/rust-mbedtls/issues/208 (making https://github.com/fortanix/rust-mbedtls/pull/209 obsolete) while bringing `rust-mbedtls` to 32 bit targets (specifically the ARM `thumb...` targets commonly found on MCUs).

With the changes presented in this PR, I was finally able to successfully use `rust-mbedtls` on an STM32L452CCUx MCU.

### DIFF
--- a/mbedtls/src/cipher/raw/serde.rs
+++ b/mbedtls/src/cipher/raw/serde.rs
@@ -341,20 +341,20 @@ unsafe impl BytesSerde for gcm_context {}
 
 // If the C API changes, the serde implementation needs to be reviewed for correctness.
 
-unsafe fn _check_cipher_context_t_size(ctx: cipher_context_t) -> [u8; 96] {
+unsafe fn _check_cipher_context_t_size(ctx: cipher_context_t) -> [u8; size_of::<cipher_context_t>()] {
     ::core::mem::transmute(ctx)
 }
 
-unsafe fn _check_aes_context_size(ctx: aes_context) -> [u8; 288] {
+unsafe fn _check_aes_context_size(ctx: aes_context) -> [u8; size_of::<aes_context>()] {
     ::core::mem::transmute(ctx)
 }
 
-unsafe fn _check_des_context_size(ctx: des_context) -> [u8; 128] {
+unsafe fn _check_des_context_size(ctx: des_context) -> [u8; size_of::<des_context>()] {
     ::core::mem::transmute(ctx)
 }
 
-unsafe fn _check_des3_context_size(ctx: des3_context) -> [u8; 384] {
+unsafe fn _check_des3_context_size(ctx: des3_context) -> [u8; size_of::<des3_context>()] {
     ::core::mem::transmute(ctx)
 }
 
-unsafe fn _check_gcm_context_size(ctx: gcm_context) -> [u8; 424] { ::core::mem::transmute(ctx) }
+unsafe fn _check_gcm_context_size(ctx: gcm_context) -> [u8; size_of::<gcm_context>()] { ::core::mem::transmute(ctx) }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,15 +197,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "core_io"
-version = "0.1.20210325"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "97f8932064288cc79feb4d343a399d353a6f6f001e586ece47fe518a9e8507df"
-dependencies = [
- "rustc_version",
-]
-
-[[package]]
 name = "env_logger"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -377,7 +368,6 @@ dependencies = [
  "cc",
  "cfg-if 1.0.0",
  "chrono",
- "core_io",
  "hex",
  "hyper",
  "libc",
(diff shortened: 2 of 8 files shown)


DrTobe:
@zugzwang Would you mind having a look at this or find another maintainer who could find the time?

zugzwang:
Hello @DrTobe, thanks for the PR!
We will shortly be in touch, please allow some days.

mbedtls/src/cipher/raw/serde.rs
```diff
@@ -341,20 +341,20 @@ unsafe impl BytesSerde for gcm_context {}
 
 // If the C API changes, the serde implementation needs to be reviewed for correctness.
 
-unsafe fn _check_cipher_context_t_size(ctx: cipher_context_t) -> [u8; 96] {
+unsafe fn _check_cipher_context_t_size(ctx: cipher_context_t) -> [u8; size_of::<cipher_context_t>()] {
```
zugzwang, line 344: I think we need to use constants here, as compilation of these functions alerts us of changes in these objects.

E.g. if for some reason `cipher_context_t` changes size, then this will result in a compilation failure. However using `cipher_context_t` against its own size will always work. 

Does this make sense?

DrTobe: Ah yes, I see, this definitely makes sense. But the value of that constant must be platform-dependent. So, it should roughly be `x + y * size_of::<usize>()` I guess?

zugzwang: I don't think so, all these structs fields are only dependent on crypto (see e.g. https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/cipher.h#L319 for `cipher_context_t`). E.g. any platform should serialize a `cipher_context_t` in exactly 96 bytes.

(Orthogonally, I do think that these constants should be declared separately e.g. `const CIPHER_CONTEXT_T_SIZE: usize = 96;`)

DrTobe: Okay, I haven't checked this specific check in detail but I have changed it because some of the checks failed when building for `thumbv7em-none-eabihf` so I just assumed now this would be due to different `usize` sizes. I will take a look at this again soon.

zugzwang: Oh, different _usize_ sizes? I thought `usize` is always 8 bits.
You are right, we should adjust by target (and perhaps add some sanity check test that checks that serialized objects are really exactly 96, 288, 128, etc bytes, if possible)

DrTobe: `usize` is _the_ type that comes to my mind when talking about different sizes in memory because it is
> The pointer-sized unsigned integer type.
> 
> The size of this primitive is how many bytes it takes to reference any location in memory. For example, on a 32 bit target, this is 4 bytes and on a 64 bit target, this is 8 bytes.

See https://doc.rust-lang.org/std/primitive.usize.html

zugzwang: Thanks, so let us adjust those constants according to the target.

DrTobe: I had a look at the underlying C structs and calculated the expected sizes. On a 64 bit machine, this works. I still have to test this when building for a 32 bit target.

DrTobe: With the latest commit, this also works for 32 bit targets 👍

jyao1:
Hello, this seems good solution to replace #210. 
We do need make it work in no_std environment with latest rust toolchain.
Any chance to merge ?

jyao1:
@MihirLuthra, any change to merge this one?

mbedtls/src/cipher/raw/serde.rs
```diff
@@ -339,22 +339,34 @@ unsafe impl BytesSerde for des_context {}
 unsafe impl BytesSerde for des3_context {}
 unsafe impl BytesSerde for gcm_context {}
 
-// If the C API changes, the serde implementation needs to be reviewed for correctness.
+// If the C API changes, the serde implementation needs to be reviewed for correctness. The
+// following (unused) functions will most probably fail to compile when this happens so a
+// compilation failure here reminds us of reviewing the serde impl.
 
-unsafe fn _check_cipher_context_t_size(ctx: cipher_context_t) -> [u8; 96] {
+// The sizes of usize and isize as well as all pointer types will be dependent on the architecture
```
zugzwang, line 346: Thanks, this looks good now.

zugzwang (approved):
Thanks for the PR. Looks good to me.

zugzwang:
@DrTobe do you think that we can add tests for these changes? (apart from the adapted the `client_server` test?)

jyao1:
Ping!

What is the status of this PR?

DrTobe:
> Ping!
> 
> What is the status of this PR?

Sorry, I have not found the time yet to look into @zugzwang's request to add tests.

DrTobe:
> @DrTobe do you think that we can add tests for these changes? (apart from the adapted the `client_server` test?)

I have had a look at this again and to me, the adaption in the `client_server` test looks just right because it exactly tests the changes made in this PR. What other test are you imagining?

zugzwang:
@DrTobe it was more of a question in case you thought more tests could be added.
Thanks for the PR!

DrTobe:
The tests failed because the `verify_chain` test uses certificates which expired last december. I have included the callback which removes the `CERT_EXPIRED` flag to every verification attempt. This seems to be the right thing to do because the test results should not depend on the current time.

Additionally, the test did the same checks twice. I have removed that duplication.

@zugzwang Assuming that the checks succeed now, can you try to merge again?

PR #112, opened 2020-06-12
Support for certificate verification with revocation lists

raoulstrackx:
The `LinkedCertificate::verify` function was modified to also optionally accept a CRL.

As this changes the API, the version is bumped

PR #214, opened 2022-11-07
Add support for certificate revocation lists

monokles:
Originally introduced by PR #112 but never made it into master. 
I cherry-picked the original commits onto a fresh branch and fixed the resulting code.

Since it changes the API, the version number is bumped.


mbedtls/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mbedtls"
-version = "0.8.2"
+version = "0.8.3"
```
raoulstrackx, line 3: You changed the interface, so these versions are not compatible with one another.
When you only update the patch version, and dependencies are defined as `mbedtls = 0.8`, Cargo may automatically pick up the new version. Unfortunately that will lead to compile issues.
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

mbedtls/src/x509/certificate.rs
```diff
@@ -1443,4 +1446,44 @@ cYp0bH/RcPTC0Z+ZaqSWMtfxRrk63MJQF9EXpDCdvQRcTMD9D85DJrMKn8aumq0M
         assert!(crate::tests::TestTrait::<dyn Sync, MbedtlsBox<Certificate>>::new().impls_trait(), "MbedtlsBox<Certificate> should be Sync");
         assert!(crate::tests::TestTrait::<dyn Sync, MbedtlsList<Certificate>>::new().impls_trait(), "MbedtlsList<Certificate> should be Sync");
     }
+
+    #[test]
+    fn empty_crl_test() {
```
raoulstrackx, line 1451: Nitpick: Can you squash the three commits related to tests?

PR #222, opened 2023-03-01
Update expired certs for testing

Taowyoo:


Taowyoo:
Closed since error cannot be reproduced on local and remote on master now

zugzwang:
@Taowyoo this is not a reason for closure; we need to know how and when was this fixed.

This was addressed by @DrTobe in #211. 

Thanks again @DrTobe for that PR!

PR #224, opened 2023-03-09
Merge async support branch to 0.7 rust-mbedtls

Taowyoo:
This is the successor PR for https://github.com/fortanix/rust-mbedtls/pull/127 but with some more latest changes needed

PR #225, opened 2023-03-13
Update to mbedtls-sys 2.28, update test keys, and make sure async tests are run

zugzwang:


zugzwang:
bors try

Taowyoo:
bors try

Taowyoo:
bors try-

Taowyoo:
bors try

Taowyoo:
bors try

Taowyoo:
bors try

Taowyoo:
bors try

Taowyoo:
bors try

Taowyoo:
bors try-

Taowyoo:
bors try

Taowyoo:
bors try

Taowyoo:
bors try

.travis.yml
Taowyoo, line 1: The new CI config is follow the master version's way

ct.sh
Taowyoo, line 1: Same to `.travis.yml` The new CI config is follow the master version's way

Taowyoo (approved):

Taowyoo:
bors try

mbedtls/src/bignum/mod.rs
```diff
@@ -173,7 +173,7 @@ impl Mpi {
             mpi_write_string(
                 &self.inner,
                 radix,
-                buf.as_mut_ptr() as *mut i8,
+                buf.as_mut_ptr() as *mut _,
```
zugzwang, line 176: By curiosity, why is this needed?

Taowyoo: OK Let me check. I sync these few lines of code  with master version
This `i8` to `_` changes may be not necessary here. 
When I first testing, I used the a copy of master version ct.sh which will also testing on aarch64 platform. So in aarch64, it need to be u8.
So `_` could be automatically infer to i8 or 8u based on different architecture.

Although 0.7 does not include support `aarch64`.  If you want I can reverse this modification.
I feel it's fine to just leave change here.

mbedtls/src/pkcs12/mod.rs
```diff
@@ -356,6 +356,7 @@ impl BERDecodable for EncryptedData {
 
 // EncryptedContentInfo from PKCS7 see RFC 2315 section 10.1
 #[derive(Debug, Clone)]
+#[allow(dead_code)]
```
zugzwang, line 359: I am not fond of these changes.

If this needed for feature compilation? In that case the struct should be behind a feature. 
Is it needed just to get rid of the compiler warning, but later on the field is reintroduced and used?

In both cases, don't fix now so we get this unblocked, but could we open an issue to remove `allow(dead_code)` after we finish with all of this?

mbedtls/Cargo.toml
```diff
@@ -31,6 +31,7 @@ bit-vec = { version = "0.5", optional = true }
 block-modes = { version = "0.3", optional = true }
 rc2 = { version = "0.3", optional = true }
 tokio = { version = "0.3", optional = true }
+proc-macro2 = "=1.0.43"
```
zugzwang, line 34: I see that current master has `proc-macro2` versions 1.0.24 and 0.4.30 (see Cargo.lock in master), perhaps we could use one of these instead, if it helps?

zugzwang (commented):
These commits look good to me, although
- Please address comments
- I cannot approve the PR since I am an author.

mbedtls/src/pkcs12/mod.rs
Taowyoo: Sure. This is intoduced when `cargo test --features pkcs12` these two struct become dead code and thus blocked by `#![deny(warnings)]` in `src/lib.rs`

mbedtls/Cargo.toml
Taowyoo: Yes, actually I could reverse the change of:
`proc-macro2 = "=1.0.43"` and `core_io = { version = "0.1.20210325"`
They are used for updateing lock file to resolve deps problem.
But I think it's better to retain them here to give more context info.
Let me update them to sync with master

mbedtls/src/pkcs12/mod.rs
Taowyoo: Will create a issue for this after merging the #225 & #127

Taowyoo: We could leave these `#[allow(dead_code)]` in #225 and #127
And I could try to remove them in #224

mbedtls/src/bignum/mod.rs
zugzwang: It's fine, master has the same. Thanks.

mbedtls/Cargo.toml
zugzwang: Sure, as long as we don't forget. As per core_io, it's fine, it's removed by in #211 later on.

ct.sh
```diff
@@ -9,28 +9,41 @@ if [ -z $TRAVIS_RUST_VERSION ]; then
...
+        # make sure that explicitly providing the default target works
+        cargo test --target $TARGET --release
+        cargo test --features pkcs12 --target $TARGET
+        cargo test --features pkcs12_rc2 --target $TARGET
+        cargo test --features dsa --target $TARGET
+        cargo test --features spin_threading --target $TARGET
+        cargo test --features rust_threading --target $TARGET
+        cargo test --features custom_time,custom_gmtime_r --target $TARGET
+        # test the async support
+        cargo test --test async_session --features=std,threading,tokio,tokio/net,tokio/io-util,tokio/macros,tokio/rt --target $TARGET
```
mzohreva, line 29: all the tokio features can be replaced with `tokio/full`

mzohreva:
Looks good to me. I'll leave it up to you how you want to merge things.

ct.sh
Taowyoo: nice catch!

PR #234, opened 2023-03-20
Make mbedtls windows ready, force CLANG on sgx, rearrange CI scripts

kwakkel1000:
We have fixes in CLANG for SGX, so now we force that you use CLANG if you compile for SGX.
We can now compile mbedtls on windows (with CI checked)
CI scripts are restructured for better understanding.


mbedtls-sys/Cargo.toml
```diff
@@ -26,6 +26,9 @@ libc = { version = "0.2.0", optional = true }
 [target.'cfg(unix)'.dependencies]
 libc = { version = "0.2.0" }
 
+[target.'cfg(windows)'.dependencies]
+libc = { version = "0.2.0" }
```
Taowyoo, line 30: Just a small suggestion, you could consider merging the following two cfg blocks into one:
```suggestion
[target.'cfg(any(unix, windows))'.dependencies]
libc = { version = "0.2.0" }
```

Taowyoo (approved):
LGTM

ci.sh
```diff
@@ -0,0 +1,56 @@
...
+cd "$(dirname "$0")"
+
+repo_root=$(readlink -f $(dirname "${BASH_SOURCE[0]}"))
+
+if [ -z $TRAVIS_RUST_VERSION ]; then
+    echo "Expected TRAVIS_RUST_VERSION to be set in env"
+    exit 1
+fi
+
+python3 -m pip install -r ./mbedtls-sys/vendor/scripts/basic.requirements.txt
```
arai-fortanix, line 12: Can this use a python virtualenv instead of installing directly on the system?

mbedtls-platform-support/src/rust_printf.c
```diff
@@ -22,7 +28,7 @@ extern int mbedtls_printf(const char *fmt, ...) {
        return -1;
 
     n++;
-    char p[n];
+    char *p = alloca(n);
```
arai-fortanix, line 31: So this comment isn't specific to the change you're making here, since you're just changing a GNU variable-length array on the stack to an explicit `alloca()`, which is equivalent. But it's not a good idea to use either a variable-length array on the stack or an `alloca()` with an unbounded size. This allocation is on the stack, and neither the variable-length array allocation nor the `alloca()` can actually fail. This can result in stack overflow, which could result in security issues depending on the exact nature of the overflow.

There are basically two ways this could be fixed. Either setting an upper bound on the log message length (and simply truncating if the log message is longer than that), or allocating the buffer on the heap. If you're setting a maximum buffer size anad using the stack, then you may as well just use the fixed maximum size as the size of your array so you don't need `alloca()`. If you decide to heap allocate, it's probably a good idea to set a maximum size anyway, to reduce the possibility of heap exhaustion and/or attacks based on forcing a lot of heap to be allocated.

Basically: don't use `alloca()` or variable-sized arrays.

arai-fortanix: Since this is already in the code, and your change doesn't make things any worse, I'm okay with fixing the stack usage in a follow-on change.

arai-fortanix (approved):

ci.sh
Taowyoo: updated

mbedtls-platform-support/src/rust_printf.c
Taowyoo: Got, create an issue to keep tracking this : https://github.com/fortanix/rust-mbedtls/issues/317

Taowyoo:
bors ping

Taowyoo:
bors r=[arai-fortanix]

PR #235, opened 2023-03-20
Remove `core_io` related CI

Taowyoo:
Since `core_io` (which originally depends on specific version of rust toolchain: `nightly-2021-03-25`) have been removed in source code, so we need to update CI to remove unnecessary CI & tests

### DIFF
--- a/ct.sh
+++ b/ct.sh
@@ -47,9 +47,6 @@ if [ "$TRAVIS_RUST_VERSION" == "stable" ] || [ "$TRAVIS_RUST_VERSION" == "beta" 
         cargo +$TRAVIS_RUST_VERSION test --no-run --target=$TARGET
     fi
 
-elif [ $TRAVIS_RUST_VERSION = $CORE_IO_NIGHTLY ]; then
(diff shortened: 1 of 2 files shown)


Taowyoo:
@jethrogb Could you give a review on this?

ct.sh
```diff
@@ -43,13 +43,16 @@ if [ "$TRAVIS_RUST_VERSION" == "stable" ] || [ "$TRAVIS_RUST_VERSION" == "beta"
         if [ -n "$AES_NI_SUPPORT" ]; then
             cargo test --features force_aesni_support --target $TARGET
         fi
+
+        # no_std tests only be able to run on x86 platform
```
arai-fortanix, line 47: Minor: The comment should probably say "no_std tests are only able" instead of "no_std tests only be able"

arai-fortanix (approved):

ct.sh
Taowyoo: thx for point out
fixed

Taowyoo:
bors merge

PR #236, opened 2023-03-20
[v0.7] Remove pining of indirect dependencies

Taowyoo:
Because  `core_io` still depends on `nightly-20210325`, pining the  indirect dependencies of `core_io` makes this crate incompatible with a lot of crates which is on `stable` or `nightly` channel.

This PR remove the explicitly pining of  indirect dependencies to resolve above problem and keep `lock` file unchanged so the tests & CI could still pass

### DIFF
--- a/mbedtls/Cargo.toml
+++ b/mbedtls/Cargo.toml
@@ -31,6 +31,10 @@ bit-vec = { version = "0.5", optional = true }
 cbc = { version = "0.1.2", optional = true }
 rc2 = { version = "0.8.1", optional = true }
 tokio = { version = "1.16.1", optional = true }
+# Note: following two lines of pinning is necessary when you want to use `core-io` feature
+# proc-macro2 = { version = ">=1.0.24, <1.0.40" }
+# quote = { version = "=1.0.10" }
+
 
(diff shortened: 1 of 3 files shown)


mbedtls/Cargo.toml
```diff
@@ -31,8 +31,6 @@ bit-vec = { version = "0.5", optional = true }
 cbc = { version = "0.1.2", optional = true }
 rc2 = { version = "0.8.1", optional = true }
 tokio = { version = "1.16.1", optional = true }
-proc-macro2 = { version = ">=1.0.24, <1.0.40", optional = true }
-quote = { version = "=1.0.10", optional = true }
```
mzohreva, line 35: Might be worth keeping these in commented out form with additional description of what to do in case someone needs to use the `core-io` feature, or when we want to update the `core_io` version (the fact that we probably need to pin these crate versions to get a workable Cargo.lock file)

Taowyoo: Good suggestion! Keep them commented is a good approach for who want to  use them. Will also update readme for this.

I need to remove them because these two lines will block compile in our internal project which is using newer tool-chain and some newer crates but not use the `core-io` feature.

PR #237, opened 2023-03-20
Add a new IO abstraction

jethrogb:
Implement an IO abstraction as [previously suggested](https://github.com/fortanix/rust-mbedtls/pull/211#issuecomment-1476893571) to replace Read+Write if you don't have that available. This allows you to layer any `Io` types like you would Read+Write.

cc @DrTobe hope you can live with the API change.


Taowyoo (approved):
LGTM.
Thank you for this PR for provide an better unified interface of IO. Learn a lot
It make it easier to add the async support.
Will refactor #266 based on this.  It seems I could follow same path to add maker type of Async  `AnyIo`, `Steam` and implement related traits

DrTobe:
If I understand it correctly, you did two things at once in this PR:

1. `IoCallback` + `IoCallbackUnsafe` as a safe interface

That is a nice and safe abstraction over the C interface and removes the necessity to manually implement a C callback interface. This part is a huge step forward :+1: 

2. Introducing the `Io` trait
> to layer any Io types like you would Read+Write

It took me a while to wrap my head around this but I guess it is meant to be an interface which the SSL context can wrap transparently like `Io` + `ssl::Context` = `Io` in the same way as  `Read+Write` + `ssl::Context` = `Read+Write`, right?

I see the motivation here because it would _theoretically_ allow consumers of this interface to work independently from the underlying transport layer but since this trait is defined _in this library_, this would only work for consumers which explicitly have mbedtls in mind. So it would only work for users which want to use mbedtls anyways. For this idea to work in a generic way, the trait would need to be defined outside of this library but this would just be like the start of another `core_io` library which we just got rid of. In general, I would really like to see a generic interface like that but I do not think that this belongs into `rust-mbedtls`.

Regarding the fact that there are now three similarly named traits, I think there is a loss of readibility. Additionally, I suspect that it might get harder to use `ssl::Context` now because there are multiple methods with the same name from different traits (and the private version in the `Context` itself) and it is not immediately clear what trait bounds should be used or which traits need to be implemented when. So to me, the possible advantage is less than the drawbacks here.

If you stick to the `Io` trait approach, I would like to see the "to layer any `Io` types like you would Read+Write" idea to be better documented for new users to grasp what these different traits are about.

DrTobe:
> Agreed but the ecosystem hasn't really developed something like this. Especially also considering the datagram-based abstraction.

Yes, that's right. Personally, I would not do it like this but I do not object too strongly here. So feel free to merge this.

jethrogb:
Added some more documentation on the `io` module and `Io` trait.

bors r=Taowyoo

PR #239, opened 2023-03-22
Add async IO &  async support for TLS

Taowyoo:
# TLDR
This PR added:
- Async version of IO abstraction ( only stream part, IO abstraction is introduced on PR: [Add a new IO abstraction #237](237) )
- Implementation for TLS 

# Detailed Explanation

## Background

We use `mbedtls` for crypto and SSL handshake in rust. `mbedtls::ssl` module contains most FFI wrapper for c mbedtls API.

### How `mbedtls::ssl` works

To setup a TLS client or a TLS server with `mbedtls`, mainly you need to provide two things to `mbedtls`:
- [Config](https://github.com/fortanix/rust-mbedtls/blob/c39d93a19d043587dc2346fb4c54501e71190162/mbedtls/src/ssl/config.rs#L148): a config object containing settings and parameters needed for TLS handshake
- [`BIO`][1]: similar to `openssl`, `mbedtls` need you to provide a `BIO` to it. `BIO` stands for **Basic I/O abstraction**, at here you could simply treat them as source & destination of data such as a TCP socket or file i/o. In code level, it represent as two callbacks in rust:
	```rust
	// representation of the `mbedtls_ssl_send_t` and `mbedtls_ssl_recv_t` callback function pointers.
	unsafe extern "C" fn call_recv(user_data: *mut c_void, data: *mut c_uchar, len: size_t) -> c_int where Self: Sized;
	unsafe extern "C" fn call_send(user_data: *mut c_void, data: *const c_uchar, len: size_t) -> c_int where Self: Sized;
	```
In `mbedtls`, we pass function pointers and user_data pointer to c mbedtls lib through FFI function `ssl_set_bio`.
During handshake and data transfer, the c mbedtls lib will utilize those two callbacks to send or receive data from underlying IO.

### IO abstraction

The IO abstraction PR #237 created and implemented safe rust abstraction of `BIO`:

<details>
<summary>Details</summary>

- Introduce unsafe trait [`IoCallbackUnsafe`](https://github.com/fortanix/rust-mbedtls/blob/c39d93a19d043587dc2346fb4c54501e71190162/mbedtls/src/ssl/io.rs#L36) to represent `mbedtls_ssl_send_t` and `mbedtls_ssl_recv_t` callback function pointers.
- Introduce safe trait [`IoCallback`](https://github.com/fortanix/rust-mbedtls/blob/c39d93a19d043587dc2346fb4c54501e71190162/mbedtls/src/ssl/io.rs#L48) : A safe representation of above trait. Also provides generic implementation of `IoCallback` for `IoCallbackUnsafe`. So user only need to implement this trait.
- Introduce trait [`Io`](https://github.com/fortanix/rust-mbedtls/blob/c39d93a19d043587dc2346fb4c54501e71190162/mbedtls/src/ssl/io.rs#L102) to represent a duplex socket or file descriptor that can be read from or written to in case you cannot use `std::io::Read` and `std::io::Write`.
- To integrate TCP, `std::net::TcpStream` implemented `std::io::Read` and `std::io::Write`:
	- Implemented `IoCallback` with Marker types:  `Stream` for `T: std::io::Read + std::io::Write`
	- Implemented `IoCallbackUnsafe` with Marker types:  `Stream` for `Contex<T>`
	- So you can pass `std::net::TcpStream` directly to [`mbedtls::ssl::context::Context::establish`](https://github.com/fortanix/rust-mbedtls/blob/yx/async-support_master/mbedtls/src/ssl/context.rs#L234)
- To integrate UDP, since `std::net::UdpSocket` does not implement and reader/writer trait:
	- Introduced `ConnectedUdpSocket` as wrapper of `std::net::UdpSocket` and implemented `Io` for it
	- Implemented `IoCallback` with Marker types:  `AnyIo` for `T: Io`
	- Implemented `IoCallbackUnsafe` with Marker types:  `AnyIo` for `Contex<T>`
	- So you can pass `ConnectedUdpSocket` directly to [`mbedtls::ssl::context::Context::establish`](https://github.com/fortanix/rust-mbedtls/blob/yx/async-support_master/mbedtls/src/ssl/context.rs#L234)

</details>

###  Follow same path 

To add async support, we could use `tokio::net::TcpStream` and traits `tokio::io::Read` and `tokio::io::Write` and follow the same path of what we did for TCP in [IO abstraction](###IO_abstraction):

<details>
<summary>Details</summary>

- (new) Add marker type for async Steam IO: `AsyncStream`
- (new) Add [`establish_async`](https://github.com/fortanix/rust-mbedtls/blob/yx/async-support_master/mbedtls/src/ssl/async_io.rs#L144) for `Context<T>`, this async function is implemented by wrapping the original blocking function [`handshake`](https://github.com/fortanix/rust-mbedtls/blob/yx/async-support_master/mbedtls/src/ssl/context.rs#L271) in a custom `Future` implementation.
- Implemented `IoCallback` with Marker types:  `AsyncStream` for `T: tokio::io::Read + tokio::io::Write`
- Implemented `IoCallbackUnsafe` with Marker types:  `AsyncStream` for `Contex<T>`
- So you can pass `tokio::net::TcpStream` directly to [`Context::establish_async`](https://github.com/fortanix/rust-mbedtls/blob/yx/async-support_master/mbedtls/src/ssl/async_io.rs#L144)

</details>

### Async need Context

But the key obstacle here is async code need context.
The functions needed by [`tokio::io::Read`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html) and [`tokio::io::Write` ](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html) both need a parameter: `cx` whose type is `std::task::Context`.

To address this, implementation need to adjust:
- `IoCallback` `impl`'s now are for a tuple `(&'a mut TaskContext<'b>, &'c mut T)` so that `Context` is accessible in the callback
- Introduced function [`with_bio_async`](https://github.com/fortanix/rust-mbedtls/blob/yx/async-support_master/mbedtls/src/ssl/context.rs#L199) to use a callback function to wrap the `Context` and callback functions' data pointer together and pass it to `mbedtls` through `set_bio_raw`

**Note**:
The async support for DTLS will be added in another PR: https://github.com/fortanix/rust-mbedtls/pull/241

[1]: https://www.openssl.org/docs/man1.1.1/man7/bio.html

### DIFF
--- a/mbedtls/src/ssl/async_io.rs
+++ b/mbedtls/src/ssl/async_io.rs
@@ -1,0 +1,171 @@
+/* Copyright (c) Fortanix, Inc.
+ *
+ * Licensed under the GNU General Public License, version 2 <LICENSE-GPL or
+ * https://www.gnu.org/licenses/gpl-2.0.html> or the Apache License, Version
+ * 2.0 <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0>, at your
+ * option. This file may not be copied, modified, or distributed except
+ * according to those terms. */
+
+#![cfg(all(feature = "std", feature = "async"))]
+
+use crate::{
+    error::{Error, Result},
+    ssl::{
+        context::Context,
+        io::{IoCallback, IoCallbackUnsafe},
+    },
+};
+use std::{
+    future::Future,
+    io::{Error as IoError, ErrorKind as IoErrorKind, Result as IoResult},
+    pin::Pin,
+    task::{Context as TaskContext, Poll},
+};
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+
+/// Marker type for an IO implementation that implements both
+/// `tokio::io::AsyncRead` and `tokio::io::AsyncWrite`.
+pub enum AsyncStream {}
+
+// TODO: Add enum `AnyAsyncIo` as marker type for an IO implementation that
+// doesn't implement `tokio::io::AsyncRead` and `tokio::io::AsyncWrite`.
+
+// TODO: Add `AsyncIo` trait for async IO that that doesn't implement
+// `tokio::io::AsyncRead` and `tokio::io::AsyncWrite`. For example:
+//     pub trait AsyncIo {
+//        async fn recv(&mut self, buf: &mut [u8]) -> Result<usize>;
+//         async fn send(&mut self, buf: &[u8]) -> Result<usize>;
+//     }
+// Could implement by using `async-trait` crate or
+// #![feature(async_fn_in_trait)] or Associated Types
+
+impl<'a, 'b, 'c, IO: AsyncRead + AsyncWrite + std::marker::Unpin + 'static> IoCallback<AsyncStream>
+    for (&'a mut TaskContext<'b>, &'c mut IO)
+{
+    fn recv(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let mut buf = ReadBuf::new(buf);
+        let io = Pin::new(&mut self.1);
+        match io.poll_read(self.0, &mut buf) {
+            Poll::Ready(Ok(())) => Ok(buf.filled().len()),
+            Poll::Ready(Err(_)) => Err(Error::NetRecvFailed),
+            Poll::Pending => Err(Error::SslWantRead),
+        }
+    }
+
+    fn send(&mut self, buf: &[u8]) -> Result<usize> {
+        let io = Pin::new(&mut self.1);
+        match io.poll_write(self.0, buf) {
+            Poll::Ready(Err(_)) => Err(Error::NetSendFailed),
+            Poll::Ready(Ok(n)) => Ok(n),
+            Poll::Pending => Err(Error::SslWantWrite),
+        }
+    }
+}
+
+impl<T: Unpin + AsyncRead + AsyncWrite + 'static> Context<T> {
+    pub async fn establish_async<IoType>(&mut self, io: T, hostname: Option<&str>) -> Result<()>
+    where
+        for<'c, 'cx> (&'c mut TaskContext<'cx>, &'c mut T): IoCallbackUnsafe<IoType>,
+    {
+        struct HandshakeFuture<'a, T>(&'a mut Context<T>);
+        impl<T> Future for HandshakeFuture<'_, T>
+        where
+            for<'c, 'cx> (&'c mut TaskContext<'cx>, &'c mut T): IoCallbackUnsafe<AsyncStream>,
+        {
+            type Output = Result<()>;
+            fn poll(mut self: Pin<&mut Self>, ctx: &mut TaskContext) -> std::task::Poll<Self::Output> {
+                self.0
+                    .with_bio_async(ctx, |ssl_ctx| match ssl_ctx.handshake() {
+                        Err(Error::SslWantRead) | Err(Error::SslWantWrite) => Poll::Pending,
+                        Err(e) => Poll::Ready(Err(e)),
+                        Ok(()) => Poll::Ready(Ok(())),
+                    })
+                    .unwrap_or(Poll::Ready(Err(Error::NetSendFailed)))
+            }
+        }
+
+        self.prepare_handshake(io, hostname)?;
+
+        HandshakeFuture(self).await
+    }
+}
+
+impl<T: AsyncRead> AsyncRead for Context<T>
+where
+    for<'c, 'cx> (&'c mut TaskContext<'cx>, &'c mut T): IoCallbackUnsafe<AsyncStream>,
+{
+    fn poll_read(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>, buf: &mut ReadBuf<'_>) -> Poll<IoResult<()>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        self.with_bio_async(cx, |ssl_ctx| match ssl_ctx.recv(buf.initialize_unfilled()) {
+            Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(())),
+            Err(Error::SslWantRead) => Poll::Pending,
+            Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
+            Ok(i) => {
+                buf.advance(i);
+                Poll::Ready(Ok(()))
+            }
+        })
+        .unwrap_or_else(|| Poll::Ready(Err(crate::private::error_to_io_error(Error::NetRecvFailed))))
+    }
+}
+
+impl<T: AsyncWrite + Unpin> AsyncWrite for Context<T>
+where
+    for<'c, 'cx> (&'c mut TaskContext<'cx>, &'c mut T): IoCallbackUnsafe<AsyncStream>,
+{
+    fn poll_write(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>, buf: &[u8]) -> Poll<IoResult<usize>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        self
+            .with_bio_async(cx, |ssl_ctx| match ssl_ctx.async_write(buf) {
+                Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(0)),
+                Err(Error::SslWantWrite) => Poll::Pending,
+                Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
+                Ok(i) => Poll::Ready(Ok(i)),
+            })
+            .unwrap_or_else(|| Poll::Ready(Err(crate::private::error_to_io_error(Error::NetSendFailed))))
+    }
+
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>) -> Poll<IoResult<()>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        match self
+            .with_bio_async(cx, Context::flush_output)
+            .unwrap_or(Err(Error::NetSendFailed))
+        {
+            Err(Error::SslWantWrite) => Poll::Pending,
+            Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
+            Ok(()) => Poll::Ready(Ok(())),
+        }
+    }
+
+    fn poll_shutdown(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>) -> Poll<IoResult<()>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        match self
+            .with_bio_async(cx, Context::close_notify)
+            .unwrap_or(Err(Error::NetSendFailed))
+        {
+            Err(Error::SslWantRead) | Err(Error::SslWantWrite) => Poll::Pending,
+            Err(e) => {
+                self.drop_io();
+                Poll::Ready(Err(crate::private::error_to_io_error(e)))
+            }
+            Ok(()) => {
+                self.drop_io();
+                Poll::Ready(Ok(()))
+            }
+        }
+    }
+}
+
+// TODO: AsyncIo impl for tokio::net::UdpSocket
--- a/mbedtls/src/ssl/context.rs
+++ b/mbedtls/src/ssl/context.rs
@@ -16,7 +16,7 @@ use mbedtls_sys::*;
 
 #[cfg(not(feature = "std"))]
 use crate::alloc_prelude::*;
-use crate::alloc::{List as MbedtlsList};
+use crate::alloc::List as MbedtlsList;
 use crate::error::{Error, Result, IntoResult};
 use crate::pk::Pk;
 use crate::private::UnsafeFrom;
@@ -172,35 +172,101 @@ impl<T> Context<T> {
     }
 }
 
+/// # Safety
+/// `io` must live as long as `ctx` or the next time bio is set/cleared.
+unsafe fn set_bio_raw<IoType, T: IoCallbackUnsafe<IoType>>(ctx: *mut ssl_context, io: &mut T) {
+    ssl_set_bio(
+        ctx,
+        io as *mut T as *mut c_void,
+        Some(T::call_send),
+        Some(T::call_recv),
+        None,
+    );
+}
+
+/// This function provides a way to apply async context to bio before running 
+/// any logic.
+/// Note: `bio` is a concept in common TLS implementation which refers to basic IO.
+/// openssl and mbedtls both use this concept.
+/// Ref: https://stackoverflow.com/questions/51672133/what-are-openssl-bios-how-do-they-work-how-are-bios-used-in-openssl
+#[cfg(all(feature = "std", feature = "async"))]
+impl<T> Context<T>  {
+    pub(super) fn with_bio_async<'cx, R, IoType>(&mut self, cx: &mut std::task::Context<'cx>, f: impl FnOnce(&mut Self) -> R) -> Option<R> where for<'c> (&'c mut std::task::Context<'cx>, &'c mut T): IoCallbackUnsafe<IoType> {
+        let ret;
+
+        struct BioGuard<'a, T> {
+            context: &'a mut Context<T>,
+        }
+        
+        impl<'a, T> Drop for BioGuard<'a, T> {
+            fn drop(&mut self) {
+                self.context.clear_bio();
+            }
+        }
+        // SAFETY: In the call to `set_bio_raw`, `user_data` must live as long
+        // as `ctx`, or until the bio is cleared from `ctx`. The bio is cleared
+        // at the end of this block ensured by the drop guard: [`BioGuard`]
+        unsafe {
+            // Points to `self.inner`, so safe to borrow at the same time as `self.io`
+            let ctx = self.into();
+            let mut user_data = (cx, &mut**self.io.as_mut()?);
+            set_bio_raw(ctx, &mut user_data);
+
+            let guard = BioGuard { context: self };
+
+            ret = f(guard.context);
+        }
+
+        Some(ret)
+    }
+
+    // This function is created to handle the odd behavior of `mbedtls_ssl_write()`
+    // Please check this https://github.com/Mbed-TLS/mbedtls/issues/4183 to learn more about how `mbedtls_ssl_write()` works in c-mbedtls 2.28
+    // This function ultimately ensure the semantics:
+    // Returned value `Ok(n)` always means n bytes of data has been sent into c-mbedtls's buffer (some of them might be sent out through underlying IO)
+    pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize> {
+        while self.handle().out_left > 0 {
+            self.flush_output()?;
+        }
+        // when calling `send()` here, already ensured that `ssl_context.out_left` == 0
+        match self.send(buf) {
+            // Although got `Error::SslWantWrite` means underlying IO is blocked, but some of `buf` is still saved into c-mbedtls's
+            // buffer, so we need to return size of bytes that has been buffered.
+            // Since we know before this call `out_left` was 0, all buffer (with in the MBEDTLS_SSL_OUT_CONTENT_LEN part) is buffered
+            Err(Error::SslWantWrite) => Ok(std::cmp::min(unsafe { ssl_get_max_out_record_payload((&*self).into()).into_result()? as usize }, buf.len())),
+            res => res,
+        }
+    }
+}
+
 impl<T> Context<T> {
     /// Establish a TLS session on the given `io`.
     ///
-    /// Upon succesful return, the context can be communicated with using the
+    /// Upon successful return, the context can be communicated with using the
     /// `std::io::Read` and `std::io::Write` traits if `io` implements those as
     /// well, and using the `mbedtls::ssl::io::Io` trait otherwise.
     pub fn establish<IoType>(&mut self, io: T, hostname: Option<&str>) -> Result<()> where T: IoCallbackUnsafe<IoType> {
+        // SAFETY: In the call to `set_bio_raw`, `self.io` must live as long as
+        // `self`, or until the bio is cleared from `ctx`. It lives as long as
+        // `self` since it is stored in self and never cleared.
         unsafe {
-            let mut io = Box::new(io);
+            self.prepare_handshake(io, hostname)?;
+            set_bio_raw(self.into(), &mut**self.io.as_mut().unwrap());
+        }
+        self.handshake()
+    }
+
+    pub(super) fn prepare_handshake(&mut self, io: T, hostname: Option<&str>) -> Result<()> {
+        unsafe {
             ssl_session_reset(self.into()).into_result()?;
             self.set_hostname(hostname)?;
             if let Some(client_id) = self.client_transport_id.take() {
                 self.set_client_transport_id(&client_id)?;
             }
-
-            let ptr = &mut *io as *mut _ as *mut c_void;
-            ssl_set_bio(
-                self.into(),
-                ptr,
-                Some(T::call_send),
-                Some(T::call_recv),
-                None,
-            );
-
-            self.io = Some(io);
+            self.io = Some(Box::new(io));
             self.inner.reset_handshake();
+            Ok(())
         }
-
-        self.handshake()
     }
 }
 
@@ -212,7 +278,7 @@ impl<T> Context<T> {
     ///
     /// This should only be used directly if the handshake could not be completed successfully in
     /// `establish`, i.e.:
-    /// - If using nonblocking operation and `establish` failed with [`Error::SslWantRead`] or
+    /// - If using non-blocking operation and `establish` failed with [`Error::SslWantRead`] or
     /// [`Error::SslWantWrite`]
     /// - If running a DTLS server and it answers the first `ClientHello` (without cookie) with a
     /// `HelloVerifyRequest`, i.e. `establish` failed with [`Error::SslHelloVerifyRequired`]
@@ -246,9 +312,16 @@ impl<T> Context<T> {
     }
 
     fn inner_handshake(&mut self) -> Result<()> {
+        self.flush_output()?;
         unsafe {
-            ssl_flush_output(self.into()).into_result()?;
             ssl_handshake(self.into()).into_result_discard()
+        }
+    }
+
+    pub(super) fn flush_output(&mut self) -> Result<()> {
+        unsafe {
+            // non-negative return value just means `ssl_flush_output` is succeed
+            ssl_flush_output(self.into()).into_result_discard()
         }
     }
 
@@ -284,15 +357,32 @@ impl<T> Context<T> {
     pub fn config(&self) -> &Arc<Config> {
         &self.config
     }
-    
-    pub fn close(&mut self) {
+
+    pub(super) fn close_notify(&mut self) -> Result<()> {
         unsafe {
-            ssl_close_notify(self.into());
+            ssl_close_notify(self.into()).into_result().map(|_| ())
+        }
+    }
+
+    pub fn close(&mut self) {
+        let _ = self.close_notify();
+        self.drop_io();
+    }
+
+    pub(super) fn clear_bio(&mut self) {
+        // It is safe to set the bio to null using the `ssl_set_bio` function. If the bio
+        // is null, mbedtls can handle this case and will return an error if you attempt
+        // to continue using SSL after calling this function.
+        unsafe {
             ssl_set_bio(self.into(), ::core::ptr::null_mut(), None, None, None);
-            self.io = None;
         }
     }
     
+    pub(super) fn drop_io(&mut self) {
+        self.clear_bio();
+        self.io = None;
+    }
+
     pub fn io(&self) -> Option<&T> {
         self.io.as_ref().map(|v| &**v)
     }
@@ -425,7 +515,7 @@ impl<T> Drop for Context<T> {
 //
 // Class exists only during SNI callback that is configured from Config.
 // SNI Callback must provide input whose lifetime exceeds the SNI closure to avoid memory corruptions.
-// That can be achieved easily by storing certificate chains/crls inside the closure for the lifetime of the closure.
+// That can be achieved easily by storing certificate chains/CRLs inside the closure for the lifetime of the closure.
 //
 // That is due to SNI being held by an Arc inside Config.
 // Config lives longer then Context. Context lives longer then Handshake.
@@ -509,7 +599,7 @@ mod tests {
     use crate::tests::TestTrait;
     
     #[test]
-    fn handshakecontext_sync() {
+    fn handshake_context_sync() {
         assert!(!TestTrait::<dyn Sync, HandshakeContext>::new().impls_trait(), "HandshakeContext must be !Sync");
     }
 
--- a/mbedtls/tests/async_session.rs
+++ b/mbedtls/tests/async_session.rs
@@ -1,0 +1,555 @@
+/* Copyright (c) Fortanix, Inc.
+ *
+ * Licensed under the GNU General Public License, version 2 <LICENSE-GPL or
+ * https://www.gnu.org/licenses/gpl-2.0.html> or the Apache License, Version
+ * 2.0 <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0>, at your
+ * option. This file may not be copied, modified, or distributed except
+ * according to those terms. */
+
+#![cfg(not(target_env = "sgx"))]
+extern crate mbedtls;
+
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+
+use mbedtls::pk::Pk;
+use mbedtls::rng::CtrDrbg;
+use mbedtls::ssl::config::{Endpoint, Preset, Transport};
+use mbedtls::ssl::{Config, Context, Version};
+use mbedtls::x509::{Certificate, VerifyError};
+use mbedtls::Error;
+use mbedtls::Result as TlsResult;
+
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+
+mod support;
+use support::entropy::entropy_new;
+use support::keys;
+use tokio::net::TcpStream;
+
+// TODO: Add unified interface for TCP and UDP like `TransportType` in
+// `client_server.rs`
+
+async fn client(conn: TcpStream, min_version: Version, max_version: Version, exp_version: Option<Version>) -> TlsResult<()> {
+    let entropy = Arc::new(entropy_new());
+    let rng = Arc::new(CtrDrbg::new(entropy, None)?);
+    let cacert = Arc::new(Certificate::from_pem_multiple(keys::ROOT_CA_CERT.as_bytes())?);
+    let expected_flags = VerifyError::empty();
+    #[cfg(feature = "time")]
+    let expected_flags = expected_flags | VerifyError::CERT_EXPIRED;
+    {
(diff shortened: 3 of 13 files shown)


mbedtls/src/ssl/context.rs
```diff
@@ -500,6 +511,250 @@ impl HandshakeContext {
...
+            io.write_tracker.adjust_buf(buf)
+        }?;
+
+        self.io_mut()
+            .ok_or(IoError::new(IoErrorKind::Other, "stream has been shutdown"))?
+            .ecx
+            .set(cx);
+
+        let result = match unsafe {
+            ssl_write((&mut *self).into(), buf.as_ptr(), buf.len()).into_result()
```
jethrogb, line 663: Can't you implement this in terms of `self.send`? Same question for read.

mbedtls/tests/async_session.rs
```diff
@@ -0,0 +1,362 @@
...
+
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+
+mod support;
+use support::entropy::entropy_new;
+use support::keys;
+
+// TODO: Add unified interface for TCP and UDP like `TransportType` in `client_server.rs`
+
+async fn client<C: AsyncRead + AsyncWrite + Unpin + 'static, T>(
```
jethrogb, line 34: If you're only supporting TcpStream I don't see any value in having this generic.

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,266 @@
...
+pub struct ErasedContext(Rc<Cell<*mut ()>>);
+
+unsafe impl Send for ErasedContext {}
+
+impl ErasedContext {
+    pub fn new() -> Self {
+        Self(Rc::new(Cell::new(null_mut())))
+    }
+
+    pub unsafe fn get(&self) -> Option<&mut TaskContext<'_>> {
```
jethrogb, line 44: Please document what the caller needs to ensure to make this safe.

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,266 @@
...
+    result::Result as StdResult,
+    task::{Context as TaskContext, Poll},
+};
+use tokio::{
+    io::{AsyncRead, AsyncWrite, ReadBuf},
+    net::UdpSocket,
+};
+
+#[derive(Clone)]
+pub struct ErasedContext(Rc<Cell<*mut ()>>);
```
jethrogb, line 35: What does this do?

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+pub struct AsyncIoAdapter<S> {
+    pub inner: S,
+    pub ecx: ErasedContext,
+    pub write_tracker: WriteTracker,
+}
+
+impl<S> AsyncIoAdapter<S> {
+    pub fn new(io: S) -> Self {
```
jethrogb, line 147: Should this be pub?

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            Err(Error::NetSendFailed)
+        }
+    }
+}
+
+/// A `tokio::net::UdpSocket` on which `connect` was successfully called.
+///
+/// Construct this type using `ConnectedAsyncUdpSocket::connect`.
+pub struct ConnectedAsyncUdpSocket {
```
jethrogb, line 232: This seems to be untested. Either remove it or implement tests.

mbedtls/src/ssl/context.rs
```diff
@@ -261,7 +272,7 @@ impl<T> Context<T> {
     }
 
     #[cfg(feature = "std")]
-    fn set_hostname(&mut self, hostname: Option<&str>) -> Result<()> {
+    pub fn set_hostname(&mut self, hostname: Option<&str>) -> Result<()> {
```
jethrogb, line 275: Why is this needed?

mbedtls/src/ssl/context.rs
```diff
@@ -500,6 +511,250 @@ impl HandshakeContext {
...
+            .establish_async(io, hostname)
+            .await
+            .map_err(|e| crate::private::error_to_io_error(e))?;
+        Ok(context)
+    }
+
+    pub async fn establish_async<IoType>(&mut self, io: T, hostname: Option<&str>) -> Result<()>
+    where
+        AsyncIoAdapter<T>: IoCallbackUnsafe<IoType>,
+    {
```
jethrogb, line 560: I feel like this should be implementable in terms of Context::establish

mbedtls/src/ssl/context.rs
```diff
@@ -500,6 +511,250 @@ impl HandshakeContext {
...
+            self.io = Some(io);
+            self.inner.reset_handshake();
+        }
+
+        HandshakeFuture(self).await
+    }
+}
+
+#[cfg(all(feature = "std", feature = "async"))]
+impl<T: AsyncRead> AsyncRead for Context<AsyncIoAdapter<T>>
```
jethrogb, line 585: Please move to async_io.rs (like in io.rs). Same for write.

mbedtls/src/ssl/context.rs
```diff
@@ -500,6 +511,250 @@ impl HandshakeContext {
...
+                unsafe {
+                    ssl_set_bio(
+                        (&mut *self).into(),
+                        ::core::ptr::null_mut(),
+                        None,
+                        None,
+                        None,
+                    );
+                }
+                self.io = None;
```
jethrogb, line 719: This code is repeated 3×, please find a way to consolidate.

Taowyoo: refactored

mbedtls/src/ssl/context.rs
Taowyoo: moved

mbedtls/src/ssl/context.rs
Taowyoo: yes, it's possible. but need some more refactor of Context::establish
updated

mbedtls/src/ssl/context.rs
Taowyoo: cleaned

mbedtls/src/ssl/async_io.rs
Taowyoo: removed
will put all UDP logic into another PR. but I still not completely figure how to do that

mbedtls/src/ssl/async_io.rs
Taowyoo: removed

mbedtls/src/ssl/async_io.rs
Taowyoo: Added

mbedtls/tests/async_session.rs
Taowyoo: removed

mbedtls/src/ssl/context.rs
Taowyoo: fixed

mbedtls/src/ssl/async_io.rs
Taowyoo: Added comments

Taowyoo: This type is a wrapper for arbitrary [`std::task::Context`]. It's used with
[`AsyncIoAdapter`] to pass the context of current async task into [`IoCallback`]
so that async context could be access in the callback inside the `mbedtls`

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +177,68 @@ impl<T> Context<T> {
...
+}
+
+#[cfg(all(feature = "std", feature = "async"))]
+impl<T> Context<T>  {
+    pub(super) fn with_bio_async<'cx, R, IoType>(&mut self, cx: &mut std::task::Context<'cx>, f: impl FnOnce(&mut Self) -> R) -> Option<R> where for<'c> (&'c mut std::task::Context<'cx>, &'c mut T): IoCallbackUnsafe<IoType> {
+        let ret;
+
+        // SAFETY: In the call to `set_bio_raw`, `user_data` must live as long
+        // as `ctx`, or until the bio is cleared from `ctx`. The bio is cleared
+        // at the end of this block.
```
jethrogb, line 199: We should probably have a drop guard for clear_bio

mbedtls/src/ssl/context.rs
jethrogb: Still shows as pub?

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,284 @@
...
+                Err(Error::SslWantRead) | Err(Error::SslWantWrite) => Poll::Pending,
+                Err(e) => Poll::Ready(Err(e)),
+                Ok(()) => Poll::Ready(Ok(())),
+            })
+            .unwrap_or(Poll::Ready(Err(Error::NetSendFailed)))
+    }
+}
+
+impl<T: Unpin + AsyncRead + AsyncWrite + 'static> Context<T> {
+    pub async fn accept_async<IoType>(config: Arc<Config>, io: T, hostname: Option<&str>) -> IoResult<Context<T>>
```
jethrogb, line 181: We don't have any special constructors anymore, I think we can get rid of this now. Caller can do this.

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,284 @@
...
+        let io = Pin::new(&mut self.1);
+        match io.poll_write(self.0, buf) {
+            Poll::Ready(Ok(n)) => Ok(n),
+            Poll::Ready(Err(_)) => Err(Error::NetSendFailed),
+            Poll::Pending => Err(Error::SslWantWrite),
+        }
+    }
+}
+
+struct HandshakeFuture<'a, T>(&'a mut Context<T>);
```
jethrogb, line 162: Please move this definition into `fn establish_async`

mbedtls/src/ssl/context.rs
Taowyoo: Agree.
Added

mbedtls/src/ssl/async_io.rs
Taowyoo: Yeah. Removed

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,253 @@
...
+    for<'c, 'cx> (&'c mut TaskContext<'cx>, &'c mut T): IoCallbackUnsafe<AsyncStream>,
+{
+    fn poll_write(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>, buf: &[u8]) -> Poll<IoResult<usize>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        let result = self
+            .with_bio_async(cx, |ssl_ctx| {
+                ssl_ctx.write_tracker.adjust_buf(buf)?;
```
jjfiv, line 204: This ``write_tracker`` call advances ``buf`` by returning it; why aren't you capturing that return here? Document with a comment if OK to discard.

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +177,77 @@ impl<T> Context<T> {
...
+        io as *mut T as *mut c_void,
+        Some(T::call_send),
+        Some(T::call_recv),
+        None,
+    );
+}
+
+#[cfg(all(feature = "std", feature = "async"))]
+impl<T> Context<T>  {
+    pub(super) fn with_bio_async<'cx, R, IoType>(&mut self, cx: &mut std::task::Context<'cx>, f: impl FnOnce(&mut Self) -> R) -> Option<R> where for<'c> (&'c mut std::task::Context<'cx>, &'c mut T): IoCallbackUnsafe<IoType> {
```
jjfiv, line 194: This must be a TLS thing, but I feel like somewhere there should be a clue as to what "BIO" stands for in one of the comments... or is it just like, buffered-I/O?

If it's just buffered I/O, then why not ``with_buf_io_async``?

mbedtls/src/ssl/context.rs
```diff
@@ -284,15 +331,29 @@ impl<T> Context<T> {
...
+            ssl_close_notify(self.into()).into_result().map(|_| ())
+        }
+    }
+
     pub fn close(&mut self) {
+        let _ = self.close_notify();
+        self.drop_io();
+    }
+
+    pub(super) fn clear_bio(&mut self) {
```
jjfiv, line 346: Mark this function as ``unsafe`` if the whole body is ``unsafe``; add SAFETY comment here as well.

mbedtls/src/ssl/context.rs
Taowyoo: `bio` is a concept in common TLS implementation which refers to basic IO.
openssl and mbedtls both use this concept: 
Ref: https://stackoverflow.com/questions/51672133/what-are-openssl-bios-how-do-they-work-how-are-bios-used-in-openssl

I will update comments for explaining this

mbedtls/src/ssl/context.rs
Taowyoo: Nice catch!

jjfiv: If it’s mostly unsafe because of FFI it could be better the other way, it depends.

mbedtls/src/ssl/async_io.rs
Taowyoo: fixed

mbedtls/src/ssl/context.rs
jethrogb: I don't think this is unsafe.

Taowyoo: Same idea, I have add comments about why it's safe

jethrogb: Should be a regular comment, not a doc comment.

jethrogb: Should be a regular comment, not a doc comment. Safety information is normally put on the `unsafe` block and not on the containing function.

Taowyoo: Got it

Taowyoo: fix it

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+}
+
+impl<T: Unpin + AsyncRead + AsyncWrite + 'static> Context<T> {
+    pub async fn establish_async<IoType>(&mut self, io: T, hostname: Option<&str>) -> Result<()>
+    where
+        for<'c, 'cx> (&'c mut TaskContext<'cx>, &'c mut T): IoCallbackUnsafe<IoType>,
+    {
+        self.prepare_handshake(io, hostname)?;
```
jethrogb, line 148: Nit: normally, function-local item definitions are placed at the top, so the code isn't interlaced with them.

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,253 @@
...
+        stream.poll_flush(cx)
+    }
+
+    fn poll_shutdown(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>) -> Poll<IoResult<()>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        match self
+            .with_bio_async(cx, Context::close_notify)
```
jethrogb, line 237: Are you allowed to call `close_notify` if `ssl_write` most recently returned `MBEDTLS_ERR_SSL_WANT_WRITE/READ` (i.e. if WriteTracker has pending data)?

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,253 @@
...
+                match ssl_ctx.send(buf) {
+                    Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(0)),
+                    Err(Error::SslWantWrite) => Poll::Pending,
+                    Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
+                    Ok(i) => Poll::Ready(Ok(i)),
+                }
+            })
+            .unwrap_or_else(|| Poll::Ready(Err(crate::private::error_to_io_error(Error::NetSendFailed))));
+
+        self.write_tracker.post_write(buf, &result);
```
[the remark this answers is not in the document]
Taowyoo: Em, it's little odd here. I just copy paste the `WriteTracker` from Mohsen's original code. 
It seems the `WriteTracker` only track the length of buff used in last call.

And when first time pending is returned, `post_write()` always is first one to record the length of buf, and in next call to `poll_write`, adjust_buf() will adjust the length based on it's internal state

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        cx.waker().clone().wake();
+
+        result
+    }
+
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>) -> Poll<IoResult<()>> {
+        // We can only flush the actual IO here.
+        // To flush mbedtls we need writes with the same buffer until complete.
```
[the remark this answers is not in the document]
Taowyoo: Based on the comments in the thread above, it appears that the `pending` status returned by c-mbedtls, which is then stored in `WriteTracker`, simply indicates that there is data waiting to be sent through the underlying IO. As a result, calling `stream.poll_flush(cx)` should effectively flush this data.

mbedtls/src/ssl/async_io.rs
Taowyoo: According to C code, it seems only `MBEDTLS_ERR_SSL_WANT_WRITE` affect `close_notify` behavior: it change to call and return `mbedtls_ssl_flush_output` instead.
So the above code should be okay :confused:
https://github.com/fortanix/rust-mbedtls/blob/fceaf689f879cc73c75393ad7b45ba238f7d6776/mbedtls-sys/vendor/library/ssl_msg.c#L5784-L5810

mbedtls/src/ssl/async_io.rs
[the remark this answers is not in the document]
jethrogb: I'm not sure what you're trying to say. I think the WriteTracker implementation is correct. All I'm saying is that you're not using WriteTracker correctly. The implementation in Mohsen's PR is correct and doesn't have the issue I pointed out here.

mbedtls/src/ssl/async_io.rs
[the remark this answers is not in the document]
jethrogb: The problem is that MbedTLS may have buffered some data internally. That data would normally be sent on the next call to `ssl_write`, but it's not possible to call `ssl_write` if there is pending data. This is not addressed by `stream.poll_flush`.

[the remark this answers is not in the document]
jethrogb: There's a function `ssl_flush_output`, can you investigate if we could use that, and whether that's callable if WriteTracker has pending data?

jethrogb:
The WriteTracker refactor could be more optimized using:
```
fn pending(&mut self) -> &mut Option<DigestAndLen> {
    self.pending.get_or_insert_with(|| Box::new(None))
}
```
However, after looking a bit more into the MbedTLS write code, I think we should get rid of WriteTracker all together. The MbedTLS requirement that ssl_write be called with the same arguments appears to be too strictly worded. MbedTLS uses internal buffering, so you just need to make sure that the internal buffer is cleared before writing new data. I think we should change poll_write to account for MbedTLS's internal buffer. See also https://github.com/Mbed-TLS/mbedtls/issues/4183#issuecomment-1508354320

```rust
// pseudo code
fn poll_write(cx, buf) {
    if cx.out_left >0 {
        ssl_flush_output(cx)
    } else {
        match ssl_write(cx, buf.as_ptr(), buf.len()) {
            Ok(l) => Ok(l),
            WANT_WRITE if cx.out_left > 0 => Ok(cx.out_left),
            WANT_WRITE => Pending,
            e => Err(e)
        }
    }
}
```

mbedtls/src/ssl/async_io.rs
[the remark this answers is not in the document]
Taowyoo: Got it. I did not notice the MbedTLS's buffer. Will take a look to `ssl_flush_output`

Taowyoo:
## Problem of `mbedtls_ssl_write`
After go through the c-mbedtls code logic, the logic of `mbedtls_ssl_write` is:
- When `ssl->out_left != 0`, `mbedtls_ssl_write` will call `mbedtls_ssl_flush_output` and **skip `mbedtls_ssl_write()`**
- Then if `mbedtls_ssl_flush_output` succeed,  `mbedtls_ssl_write` will **return the `len` which is from current call's arguments**.

Then this makes `mbedtls_ssl_write` behaves not as expected: no new data is written into `mbedtls`'s buffer but return original buffer length indicates all data has been written. And this is why original API comments ask user to keep the arguments if `mbedtls_ssl_write()` **same**, which eliminates above behavior inconsistency.

## My proposed solution

Added function `pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize>` to address the problem with odd API behavior of `mbedtls_ssl_write`.

The function does following things:

1. If there is pending data prepared in an earlier call to `mbedtls_ssl_write()` in c-mbedtls's buffer, this function try to flush them first by calling `mbedtls_ssl_flush_output`.
2. Then call `mbedtls_ssl_write()` as normal if all pending data has been flushed out.
3. If any call above returns `Error` then just handles them as normal. Because in this case if `Error` is `SslWantWrite`, it just means the underlying IO is blocked and this function need to be called again later.

Taowyoo:
## To-do

- [ ] Add test cases to ensure above solution works

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,184 @@
...
+{
+    fn poll_write(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>, buf: &[u8]) -> Poll<IoResult<usize>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        let result = self
+            .with_bio_async(cx, |ssl_ctx| match ssl_ctx.async_write(buf) {
+                Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(0)),
+                Err(Error::SslWantWrite) => Poll::Pending,
```
jethrogb, line 136: This is not right. You need to return a succesful write for the number of bytes that were buffered internally.

Taowyoo: I see, I need to save the original `out_left` before call flush function. And return the flushed bytes number.
I put the logic of this part is at `async_write` in `context.rs`

Taowyoo: Updated `async_write` to return actual flushed number of bytes when `out_left > 0`

Taowyoo: Updated `async_write` function
Since `SslWantWrite` is only returned by the send callback in `async_io.rs`  , we know that in this case no data is written directly to underlying IO, so it could just return `Ok(out_left)`

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +172,118 @@ impl<T> Context<T> {
...
+                    Err(e) => { return Err(e); },
+                }
+            }
+            // when calling `send()` here, already ensure that `ssl_context.out_left` == 0
+            match self.send(buf) {
+                // Although got `Error::SslWantWrite` means underlying IO is blocked, but some of `buf` is still saved into c-mbedtls's
+                // buffer, so we need to return size of bytes that has been buffered
+                // We know we could only use `out_left` here because in current implementation `Error::SslWantWrite` is only return by
+                // function [`IoCallback<AsyncStream>::send`], in this case no data is written into underlying IO, which means
+                // `out_left` == size of data buffered
```
jethrogb, line 251: Add comment: Since we know out_left was 0 prior to this call.

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +172,118 @@ impl<T> Context<T> {
...
+    pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize> {
+        unsafe {
+            let inner: *mut ssl_context = self.into();
+            if (*inner).out_left > 0 {
+                match self.flush_output() {
+                    Ok(()) => {
+                        // if data prepared in an earlier call to `mbedtls_ssl_write()` is all flushed then continue to write new data
+                        if (*inner).out_left != 0 {
+                            // data prepared in an earlier call has not been flushed out
+                            // return `SslWantWrite` to caller to indicate call this function again later
```
jethrogb, line 237: I think we should keep looping until we get an error or out_left is 0. Or perhaps we need to interpret the positive return values of flush_output? In particular, we need to be sure that the inner IO has set up notifications for the current task.

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +172,118 @@ impl<T> Context<T> {
...
+    }
+
+    // This function is created to handle the old behavior of  `mbedtls_ssl_write()` only for TLS use cases
+    // If there is pending data prepared in an earlier call to `mbedtls_ssl_write()` in c-mbedtls's buffer, this function try to flush them first by calling `mbedtls_ssl_flush_output`.
+    // Then call `mbedtls_ssl_write()` as normal if all pending data has been flushed out.
+    // It handles error as normal because in case of getting `SslWantWrite` from `mbedtls_ssl_write()` or `mbedtls_ssl_flush_output()`, this function just need to be called again.
+    // Ref: https://github.com/Mbed-TLS/mbedtls/issues/4183
+    pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize> {
+        unsafe {
+            let inner: *mut ssl_context = self.into();
```
jethrogb, line 230: I don't think you need to use raw pointers? Just do `self.inner.out_left` everywhere

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,184 @@
...
+        result
+    }
+
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>) -> Poll<IoResult<()>> {
+        if self.handle().session.is_null() {
+            return Poll::Ready(Err(IoError::new(IoErrorKind::Other, "stream has been shutdown")));
+        }
+
+        match self
+            .with_bio_async(cx, |ssl_ctx| ssl_ctx.flush_output())
```
jethrogb, line 153: Nit: `self.with_bio_async(cx, Context::flush_output)`

jethrogb (commented):
Have you given some thought on how to test all this?

Taowyoo:
@jethrogb 
I tried testing with same stimulated TCP stream tools but does not work.
Si it seems to better create a test specific impl of `IoCallback`  for testing:  we need to manually control how `IoCallback`

mbedtls/src/ssl/context.rs
Taowyoo: the type of `self.inner` is `HandshakeContext` 
Could replaced them with `self.handle().out_left`

mbedtls/src/ssl/context.rs
Taowyoo: > I think we should keep looping until we get an error or out_left is 0
I could follow this approach.

Regarding:
> In particular, we need to be sure that the inner IO has set up notifications for the current task.
If just let Future's mechanism to do the retry, I understand that `cx.waker().clone().wake();` in `poll_write` already do the work for sending  notifications?

Taowyoo:
This test could verify the whether the approach of get rid of `WriteTracker` works correctly:

<details>
<summary>Test code Details</summary>

https://github.com/fortanix/rust-mbedtls/blob/40df61dbaaff31087212e4956007aa7eb12263e7/mbedtls/tests/async_session.rs#L345-L396


</details>
This test will fail on current implementation at 37550f2

Taowyoo:
I revert the recent change about removing `WriteTracker` and add a test `async_session::test::test_write_tracker` which ensures that `WriteTracker` works as expected.

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,184 @@
...
+            // Need to handle 0 return especially because:
+            // A return value of 0 typically means that the underlying object is no longer able to accept bytes and will likely
+            // not be able to in the future as well, or that the buffer provided is empty.
+            //
+            // This is important because:
+            // c-mbedtls assumes: the callback must return the number of bytes sent if any, or a non-zero error code.
+            //
+            // Ref: https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html#tymethod.poll_write
+            // Ref: https://github.com/fortanix/rust-mbedtls/blob/master/mbedtls-sys/vendor/include/mbedtls/ssl.h#L572-L591
+            Poll::Ready(Err(_)) | Poll::Ready(Ok(0)) => Err(Error::NetSendFailed),
```
jethrogb, line 67: Why is 0 handled as a special case in async only?

mbedtls/src/ssl/async_io.rs
```diff
@@ -0,0 +1,184 @@
...
+        let result = self
+            .with_bio_async(cx, |ssl_ctx| match ssl_ctx.async_write(buf) {
+                Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(0)),
+                Err(Error::SslWantWrite) => Poll::Pending,
+                Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
+                Ok(i) => Poll::Ready(Ok(i)),
+            })
+            .unwrap_or_else(|| Poll::Ready(Err(crate::private::error_to_io_error(Error::NetSendFailed))));
+
+        cx.waker().clone().wake();
```
jethrogb, line 142: This seems extremely inefficient: you're waking the task immediately every time this function is called (including if this returns Poll::Pending), basically resulting in continuous polling?

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +172,105 @@ impl<T> Context<T> {
...
+            ret = f(guard.context);
+        }
+
+        Some(ret)
+    }
+
+    // This function is created to handle the ood behavior of `mbedtls_ssl_write()` only for TLS use cases
+    // Please check this https://github.com/Mbed-TLS/mbedtls/issues/4183 to learn more about how `mbedtls_ssl_write()` works in c-mbedtls 2.28
+    // This function ultimately ensure the semitics:
+    // Returned value `Ok(n)` always means n bytes of data has been sent into c-mbedtls's buffer (some of them might be sent out through underlying IO)
```
jethrogb, line 226: What does "only for TLS use cases" mean? That seems redundant, if you have an ssl_context you're doing TLS.

Typos:
* ood -> odd
* semitics -> semantics

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +172,105 @@ impl<T> Context<T> {
...
+    pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize> {
+        // keep looping until we get an error or `out_left` is 0
+        while self.handle().out_left > 0 {
+            match self.flush_output() {
+                Ok(()) => {},
+                Err(e) => { return Err(e); },
+            }
```
jethrogb, line 233:
```suggestion
            // try operator includes the case where `Error::SslWantWrite` is returned
            self.flush_output()?;
```

mbedtls/src/ssl/context.rs
```diff
@@ -172,35 +172,105 @@ impl<T> Context<T> {
...
+
+        Some(ret)
+    }
+
+    // This function is created to handle the ood behavior of `mbedtls_ssl_write()` only for TLS use cases
+    // Please check this https://github.com/Mbed-TLS/mbedtls/issues/4183 to learn more about how `mbedtls_ssl_write()` works in c-mbedtls 2.28
+    // This function ultimately ensure the semitics:
+    // Returned value `Ok(n)` always means n bytes of data has been sent into c-mbedtls's buffer (some of them might be sent out through underlying IO)
+    pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize> {
+        // keep looping until we get an error or `out_left` is 0
```
jethrogb, line 228: Unnecessary comment

mbedtls/tests/support/net.rs
```diff
@@ -17,12 +19,47 @@ pub fn create_tcp_pair() -> IoResult<(TcpStream, TcpStream)> {
...
-                TcpStream::from_raw_fd(fds[1]),
-            ))
+            Ok((TcpStream::from_raw_fd(fds[0]), TcpStream::from_raw_fd(fds[1])))
         } else {
             Err(IoError::last_os_error())
         }
     }
 }
+
+pub fn create_tcp_stream_pair_loopback() -> (TcpStream, TcpStream) {
```
jethrogb, line 29: Why? What's wrong with `create_tcp_pair`?

mbedtls/tests/support/write_all.rs
jethrogb, line 1: Why not use `futures::AsyncWriteExt`?

mbedtls/tests/support/rand.rs
```diff
@@ -45,3 +45,16 @@ impl crate::mbedtls::rng::RngCallback for TestRandom {
 pub fn test_rng() -> TestRandom {
     TestRandom(XorShiftRng::new_unseeded())
 }
+
+/// Testing only helper function to create a buffer with random content
+#[cfg(feature = "std")]
+pub fn random_data(sz: usize) -> Vec<u8> {
```
jethrogb, line 51:
```
std::iter::repeat_with(rand::random)
    .take(sz)
    .collect()
```

mbedtls/src/ssl/async_io.rs
Taowyoo: I see, originally when I read C code of `mbedtls_ssl_flush_output`, I found it will return success code when callback return 0. So I think it might cause problem. But it seems the code there is correct.
This `0` need to be handled by out most caller

mbedtls/tests/support/net.rs
Taowyoo: Emm, it's my old code where I tried to put some latency and package loss on `loopback` interface to test.
will clean them up

mbedtls/tests/support/write_all.rs
Taowyoo: Because in `AsyncWriteExt`, those functions build up on `poll_write` actually already follows the requirements mentioned in `mbedtls_ssl_write` they do not change arguments such as `buf` when got `Pending`.

So I create a custom one to stimulate changing `buf` when got `Pending` from `poll_write`

mbedtls/src/ssl/context.rs
Taowyoo: Will update it.
What I mean here is more about this function might not work with DTLS

mbedtls/src/ssl/async_io.rs
Taowyoo: reverted to original way

mbedtls/tests/support/write_all.rs
jethrogb: Ah. This wasn't immediately clear when skimming the code. This is fine, but please use a more clear name than “write all” and also add some high level documentation about what this does.

mbedtls/tests/ssl_conf_ca_cb.rs
```diff
@@ -18,14 +18,14 @@ use mbedtls::pk::Pk;
...
-use mbedtls::x509::{Certificate};
+use mbedtls::x509::Certificate;
 use mbedtls::Result as TlsResult;
 use mbedtls::ssl::config::CaCallback;
 
 mod support;
 use support::entropy::entropy_new;
 
-use mbedtls::alloc::{List as MbedtlsList};
+use mbedtls::alloc::List as MbedtlsList;
```
jethrogb, line 28: Why is this suddenly necessary? :thinking:

jethrogb (commented):
Looks great now. Few comments in the tests and then this is ready to land I believe.

mbedtls/tests/support/write_all.rs
Taowyoo: Added comments and renamed them

Taowyoo:
bors retry

Taowyoo:
bors cancel

Taowyoo:
bors retry

Taowyoo:
bors r=[jethrogb]

PR #241, opened 2023-03-28
Add async support for DTLS

Taowyoo:
To keep things simple, the async support for DTLS is implemented by:

- Added `ConnectedAsyncUdpSocket`
- Impl `AsyncRead` and `AsyncWrite` for `ConnectedAsyncUdpSocket`
- Extend `async-session` test for testing DTLS

### DIFF
--- a/mbedtls/src/ssl/async_io.rs
+++ b/mbedtls/src/ssl/async_io.rs
@@ -125,8 +125,6 @@ impl<S> AsyncIoAdapter<S> {
     }
 }
 
-pub type AsyncContext<T> = Context<AsyncIoAdapter<T>>;
-
 /// Marker type for an IO implementation that doesn't implement
 /// `tokio::io::AsyncRead` and `tokio::io::AsyncWrite`.
 pub enum AnyAsyncIo {}
@@ -192,7 +190,7 @@ where
     fn poll(mut self: Pin<&mut Self>, ctx: &mut TaskContext) -> std::task::Poll<Self::Output> {
         self.0
             .inner
-            .with_bio_async(ctx, |sslctx| match sslctx.handshake() {
+            .with_bio_async(ctx, |ssl_ctx| match ssl_ctx.handshake() {
                 Err(Error::SslWantRead) | Err(Error::SslWantWrite) => Poll::Pending,
                 Err(e) => Poll::Ready(Err(e)),
                 Ok(()) => Poll::Ready(Ok(())),
@@ -238,7 +236,7 @@ where
         }
 
         self.inner
-            .with_bio_async(cx, |sslctx| match sslctx.recv(buf.initialize_unfilled()) {
+            .with_bio_async(cx, |ssl_ctx| match ssl_ctx.recv(buf.initialize_unfilled()) {
                 Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(())),
                 Err(Error::SslWantRead) => Poll::Pending,
                 Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
@@ -263,9 +261,9 @@ where
         let AsyncIoAdapter { inner, write_tracker } = &mut *self;
 
         let result = inner
-            .with_bio_async(cx, |sslctx| {
+            .with_bio_async(cx, |ssl_ctx| {
                 write_tracker.adjust_buf(buf);
-                match sslctx.send(buf) {
+                match ssl_ctx.send(buf) {
                     Err(Error::SslPeerCloseNotify) => Poll::Ready(Ok(0)),
                     Err(Error::SslWantWrite) => Poll::Pending,
                     Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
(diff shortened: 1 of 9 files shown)


mbedtls/tests/async_session.rs
```diff
@@ -28,17 +31,68 @@ use support::entropy::entropy_new;
...
+    }
+}
+
+#[async_trait]
+impl TransportType for ConnectedAsyncUdpSocket {
+    fn get_transport_type() -> Transport {
+        Transport::Datagram
+    }
+
+    async fn recv(ctx: &mut Context<Self>, buf: &mut [u8]) -> TlsResult<usize> {
```
Taowyoo, line 63: The `recv` and `send` here is same to `TcpStream` because I choose to also Impl `AsyncRead` and `AsyncWrite` for `ConnectedAsyncUdpSocket`

mbedtls/src/ssl/async_io.rs
```diff
@@ -250,4 +263,42 @@ where
...
+            Err(e) => Err((e, socket)),
+        }
+    }
+
+    pub fn into_socket(self) -> UdpSocket {
+        self.socket
+    }
+}
+
+impl AsyncRead for ConnectedAsyncUdpSocket {
```
Taowyoo, line 286: To reuse the code above for `TcpStream`, I choose to impl `AsyncRead` and `AsyncWrite` for `ConnectedAsyncUdpSocket`.
Although UDP does not have concept of Stream, but in API level, `AsyncRead` and `AsyncWrite` is still compatible to it.

jethrogb: Well, you could make the same argument for `Read` and `Write`. I don't think we should do this. `UdpSocket` also doesn't do this.

Taowyoo: So, you mean I still need to define a  trait with func `recv` `send` for `AnyAsyncIo` such as UdpSocket?

Taowyoo: Re-added the `AsyncIo`, integrate it with `AsyncRecv` & `AsyncSend` which are simpler version of tokio's `AsyncRead` `AsyncWrite`

mbedtls/src/ssl/async_io.rs
```diff
@@ -102,121 +118,210 @@ impl WriteTracker {
...
+/// By implementing [`AsyncSend`] and [`AsyncRecv`] you can wrap any type of
+/// [`AsyncIo`] with [`Context<T>::establish_async`] to protect that
+/// communication channel with (D)TLS. That [`Context`] then also implements
+/// [`AsyncIo`] so you can use it interchangeably.
+///
+/// If you are using byte streams and are using [`tokio::io`], you don't need
+/// this trait and can rely on [`tokio::io::AsyncRead`] and
+/// [`tokio::io::AsyncWrite`] instead.
+#[async_trait]
+pub trait AsyncIo {
```
jethrogb, line 142: Do we need both `trait AsyncIo` and [`trait AsyncSend` and `trait AsyncRecv`]? Why can't we just have one of these?

Do we need a macro for the implementations or can you just implement a private `fn async_recv`/`async_send` on a `Context` like we have for the synchronous versions?

Taowyoo: I see. I understand your main point here is reduce redundant trait and the dependency to `async-trait` crate.
So here :
It's better `trait AsyncIo` can be replaced by private functions `async_recv`/`async_send` on a `Context`

And let me try to implement `async_recv`/`async_send`  without using `async-trait`

mbedtls/src/ssl/async_io.rs
```diff
@@ -250,4 +318,42 @@ where
...
+///
+/// Construct this type using [`ConnectedAsyncUdpSocket::connect`].
+pub struct ConnectedAsyncUdpSocket {
+    socket: UdpSocket,
+}
+
+impl ConnectedAsyncUdpSocket {
+    pub async fn connect<A: tokio::net::ToSocketAddrs>(socket: UdpSocket, addr: A) -> StdResult<Self, (IoError, UdpSocket)> {
+        match socket.connect(addr).await {
+            Ok(_) => Ok(ConnectedAsyncUdpSocket { socket }),
```
jjfiv, line 332: Small nit, but for readability, I prefer to see ``Ok(())`` rather than ``Ok(_)`` because the former doesn't make me look up the signature.

mbedtls/tests/async_session.rs
```diff
@@ -51,78 +105,111 @@ async fn client(conn: TcpStream, min_version: Version, max_version: Version, exp
...
-                    }
-                    Error::SslFatalAlertMessage => {}
-                    e => panic!("Unexpected error {}", e),
-                };
-                return Ok(());
-            }
-        };
+    match ctx.establish_async(conn, None).await {
+        Ok(()) => {
+            assert_eq!(ctx.version(), exp_version.unwrap());
```
jjfiv, line 122: Not an expert, but why is this an assert which panics on error not an Err? Is this the user of this library's fault?

jjfiv: Nvm, potentially, didn't realize this was a test file.

mbedtls/src/ssl/async_io.rs
Taowyoo: I agree. will fix it

PR #244, opened 2023-04-12
Move self test logic to a separate crate

Taowyoo:
This PR is for fixing the link error: multiple define of `mbedtls_printf` when have different versions of `mbedtls` in dependency tree.

According to @jethrogb 's idea:
> I am proposing to create a new crate mbedtls-selftest which in Cargo.toml specifies links = mbedtls-selftest-support. It basically contains only the self_test module and the build script. All versions of MbedTLS would be updated to depend on a single version of mbedtls-selftest-support.

The problem could be fixed by
- moving `selftest` related implementation in `mbedtls` into a separate crate, here I named it: `mbedtls-selftest` (Done by this PR)
- back-port this PR to old versions: such as `0.7.2` `0.8.1`


mbedtls/build.rs
```diff
@@ -43,14 +43,12 @@ fn main() {
...
     b.file("src/mbedtls_malloc.c");
-    b.file("src/rust_printf.c");
     if sys_platform_components.get("c_compiler").map_or(false, |comps| comps.contains("freestanding")) {
         b.flag("-U_FORTIFY_SOURCE")
             .define("_FORTIFY_SOURCE", Some("0"))
             .flag("-ffreestanding");
     }
     b.compile("librust-mbedtls.a");
-    // Force correct link order for mbedtls_printf
     println!("cargo:rustc-link-lib=static=mbedtls");
```
jethrogb, line 52: Since you're no longer using printf in this crate, you shouldn't need these link statements anymore.

mbedtls-selftest/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[build-dependencies]
+cc = "1.0"
+
+[dependencies]
+cfg-if = "1.0.0"
+
+[features]
+# Features are documented in the README
+default = ["std"]
```
jethrogb, line 23: Probably don't need this default

mbedtls-selftest/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+edition = "2018"
+links = "mbedtls-selftest-support"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+
+[dependencies.mbedtls-sys-auto]
+version = "2.28.0"
+default-features = false
+features = ["custom_printf", "trusted_cert_callback", "threading"]
```
jethrogb, line 12: Only need `custom_printf` here, I think

Taowyoo:
I will create back-port PRs if this PR successfully merged, after back-port, new versions also need to be published
- 0.7.2 -> 0.7.3
- 0.8.2 -> 0.8.3

mbedtls/Cargo.toml
```diff
@@ -35,11 +35,15 @@ rs-libc = "0.2.0"
 chrono = "0.4"
 
 [dependencies.mbedtls-sys-auto]
-version = "2.25.0"
+version = "2.28.0"
```
jethrogb, line 38: Why are you bumping the dependency?

Taowyoo: I think, this is just a fix. The version number under ./mbedtls-sys-auto is  `2.28.0`

jethrogb: This line specifies the minimum version required. This code doesn't make any changes to MbedTLS features used, so it doesn't make sense that the requirement is changed?

Taowyoo: you're correct

mbedtls-selftest/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "mbedtls-selftest"
+version = "0.1.0"
+edition = "2018"
+links = "mbedtls-selftest-support"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+
+[dependencies.mbedtls-sys-auto]
+version = "2.28.0"
```
jethrogb, line 10: Should match requirement in `mbedtls`

Taowyoo: :+1:

mbedtls/Cargo.toml
```diff
@@ -40,6 +40,10 @@ default-features = false
 features = ["custom_printf", "trusted_cert_callback", "threading"]
```
jethrogb, line 40: I guess technically `custom_printf` is not needed here anymore.

Taowyoo: Removed

Taowyoo:
Hi, @jethrogb when testing, I found renaming `mbedtls_log` is enough to fix the link error. So what we could follow same approach you did in PR #243 

I noticed that: the link error actually is complaining about: multiple definition of `mbedtls_log' in rust code.
So actually what happened here is: c compiler what to find and link the `mbedtls_log` function implemented in `rust` into C library and found two same name function. Notice that the direction here is c -> rust, so the naming in C side does not matter:  the same naming `mbedtls_printf` in `rust_printf.c` is just similar to all other functions in C mbedtls code won't cause link name conflict error.

PR #245, opened 2023-04-13
Add suffix to `mbedtls_log` to avoid link name conflict error

Taowyoo:
## Summary of Changes in this PR

This PR addresses the link name conflict error that occurs when using different versions of the `mbedtls` crate in a single project, by adopting the approach used in PR #234.

## Rationale Behind the Approach

During testing, it was discovered that simply renaming `mbedtls_log` resolves the linking error. Therefore, the [`export_name`](https://doc.rust-lang.org/reference/abi.html#the-export_name-attribute) attribute is employed here, as done in PR #243.

Upon further investigation, it was found that the linker error stems from multiple definitions of `mbedtls_log` in the Rust code. The C compiler attempts to locate and link the `mbedtls_log` function implemented in Rust to the C library, but it encounters duplicate instances of the function with the same name.

It's important to note that the linking search direction is from C to Rust, so naming on the C side is not a concern. Functions named `mbedtls_printf` in multiple `rust_printf.c` files are similar to other functions in the C mbedtls code and do not cause link name conflict errors.

### DIFF
--- a/mbedtls/src/rust_printf.c
+++ b/mbedtls/src/rust_printf.c
@@ -9,7 +9,12 @@
 #include <stdio.h>
 #include <stdarg.h>
 
-extern void mbedtls_log(const char* msg);
+// Use several macros to get the preprocessor to actually replace RUST_MBEDTLS_METADATA_HASH
+#define append_macro_inner(a, b) a##_##b
+#define append_macro(a, b) append_macro_inner(a, b)
+#define APPEND_METADATA_HASH(f) append_macro(f, RUST_MBEDTLS_METADATA_HASH)
+
+extern void APPEND_METADATA_HASH(mbedtls_log)(const char* msg);
 
 extern int mbedtls_printf(const char *fmt, ...) {
     va_list ap;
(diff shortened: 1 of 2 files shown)


Taowyoo:
I got your point. You're correct

I had wrong understanding here
> Functions named mbedtls_printf in multiple rust_printf.c files are similar to other functions in the C mbedtls code.

Let me rephrase what happened here:

First, compiler will compile `rust_printf.c` to `rust_printf.o` where: `mbedtls_log` is Undefined symbol and `mbedtls_printf` is local defined symbol
```
# part output of `nm rust_printf.o`
                 U mbedtls_log
0000000000000000 T mbedtls_printf
```
So compiler try to find `mbedtls_log` in rust compiled objects and here is where original error occurs.
And by changing name of `mbedtls_log` , this error is resolved by this PR.

But there indeed are multiple `mbedtls_printf` in multiple rust_printf.o. By inspecting to the compiled `libmbedtls-abc.rlib` files, the the content of `rust_printf.o` is being placed to the `rlib` file correspondingly because `mbedtls` `rlib` file is compiled separately by the `build.rs` script.

So I guess the reason why compiler do not throw conflict error here is: it does not detect the usage/re-export on `mbedtls_printf`, but this is not guaranteed in runtime and the behavior on runtime will become undefined.

In short, the approach in this PR does not solve the problem correctly.

I will close this PR, and could you please review the approve the #244 if it's ok?

PR #249, opened 2023-04-17
Remove `WriteTracker` and use a simplified approach

Taowyoo:
Current implementation is not correct, the test [test_write_tracker](https://github.com/fortanix/rust-mbedtls/blob/b6e204c6aff46cceaad995e6567bea9aa219e56d/mbedtls/tests/async_session.rs#L346) cannot pass

### DIFF
--- a/mbedtls/src/ssl/context.rs
+++ b/mbedtls/src/ssl/context.rs
@@ -235,7 +235,8 @@ impl<T> Context<T>  {
         // when calling `send()` here, already ensured that `ssl_context.out_left` == 0
         match self.send(buf) {
             // Although got `Error::SslWantWrite` means underlying IO is blocked, but some of `buf` is still saved into c-mbedtls's
-            // buffer, so we need to return size of bytes that has been buffered
+            // buffer, so we need to return size of bytes that has been buffered.
+            // Since we know before this call `out_left` was 0, all buffer (with in the MBEDTLS_SSL_OUT_CONTENT_LEN part) is buffered
             Err(Error::SslWantWrite) => Ok(std::cmp::min(unsafe { ssl_get_max_out_record_payload((&*self).into()).into_result()? as usize }, buf.len())),
             ret @ _ => ret,
         }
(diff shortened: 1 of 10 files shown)


Taowyoo:
bors try

mbedtls/src/ssl/context.rs
```diff
@@ -225,6 +225,12 @@ impl<T> Context<T>  {
     // This function ultimately ensure the semitics:
     // Returned value `Ok(n)` always means n bytes of data has been sent into c-mbedtls's buffer (some of them might be sent out through underlying IO)
     pub(super) fn async_write(&mut self, buf: &[u8]) -> Result<usize> {
+        let max_len = unsafe { ssl_get_max_out_record_payload((&*self).into()).into_result()? as usize };
+        #[cfg(feature = "zlib")]
+        static RESERVE_SIZE: usize = 10;
+        #[cfg(feature = "zlib")]
+        let buf = &buf[..std::cmp::min(max_len - RESERVE_SIZE, buf.len())];
```
jethrogb: I know this is a temporary hack, but something like this is not sufficient for production. MbedTLS C code may be compiled with zlib enabled even if the Rust bindings are not.

Taowyoo: I removed this part of code.
After testing with blocking code, test also failed when caller try to write large buffer at one times.

when turn on `zlib` feature, c-mbedtls could not record buffer with size > MBEDTLS_SSL_OUT_CONTENT_LEN (default: 16 * 1024)
https://github.com/fortanix/rust-mbedtls/blob/98331fce9fcd8e9a4e472209aca182329f6de6f2/mbedtls-sys/vendor/library/ssl_msg.c#L646-L653

Disable tests for large buffer when `zlib` is enabled and add negative test case when `zlib` is enabled

Taowyoo:
bors try

Taowyoo:
bors try

Taowyoo:
Since the `zlib`  never works with large buffer, we can ignore the failing async tests when `zlib` is enabled.
So I will merge this pr into main development branch: https://github.com/fortanix/rust-mbedtls/tree/yx/async-support_master

PR #251, opened 2023-04-18
[v0.7] Move self test logic to crate `mbedtls-selftest`

Taowyoo:
# mbedtls-selftest

This Rust crate is designed for separating self-test code that needs to export Rust
functions and define C functions to be used by C `mbedtls`. By separating this code,
different versions of Rust `mbedtls` crates can be used within a single crate, which
helps to solve link name conflict errors.

**Note**: Although multiple versions of Rust `mbedtls` crates can be used within a
single crate, only one `mbedtls-selftest` and one `mbedtls-sys-auto` crate can be
used since they are built as native libraries.

## Changes need attention

- Pump version of `mbedtls` to `0.7.3`
- Remove `build.rs` in `mbedtls` since it does not contain C source file.
- Specify `crate-type = ["staticlib"]` in in `mbedtls`'s `Cargo.toml` file to
  build it into a static library.


Taowyoo:
bors try

Taowyoo:
bors try

mbedtls/Cargo.toml
```diff
@@ -39,10 +39,14 @@ tokio = { version = "1.16.1", optional = true }
 [target.x86_64-fortanix-unknown-sgx.dependencies]
 rs-libc = "0.1.0"
 
+[dependencies.mbedtls-selftest]
+version = "0.1"
+path = "../mbedtls-selftest"
```
jethrogb, line 44: Don't add it as a path dependency for the older branches

Taowyoo:
bors try

Taowyoo:
> Backport #243 as well?

0.7.X version does not have `mbedtls_malloc.c`.
Checking if there is other exported rust function that need to be placed under a namespace.

Taowyoo:
Confirmed, the 0.7.X version does not include `mbedtls_malloc.c`, so PR #243 is not necessary for this PR. 

Additionally, although there are some Rust functions that could be exported for use by C `mbedtls`, such as `mbedtls_platform_gmtime_r`. But they are not actively used and have been deprecated in version 0.8 or later, so if required in the future, we can address these functions in a separate PR.

Taowyoo:
bors try

Taowyoo:
Tested compiling mbedtls `v0.7` and `v0.8` branches as dependencies of a single crate, and they worked well, except:

Using the `custom_gmtime_r` feature in `v0.7` affects `v0.8`'s conditional compile. This enables `sys_time_component = "custom"` without the `chrono` dependency, which only is turn on when targeting SGX , causing a compilation error.

jethrogb:
But we are only setting `custom_gmtime_r` on 0.7 when compiling for SGX, no?

Taowyoo:
> But we are only setting `custom_gmtime_r` on 0.7 when compiling for SGX, no?

Yes, 👍. So we could ignore this problem.

mbedtls/src/lib.rs
```diff
@@ -56,7 +56,7 @@ pub mod ecp;
 pub mod hash;
 pub mod pk;
 pub mod rng;
-pub mod self_test;
+pub use mbedtls_selftest as self_test;
```
jethrogb, line 59: Actually this is apparently a public API change (see change in mbedtls_self_tests.rs). So we still need to have a mod self_test that keeps the API the same.

Taowyoo:
bors try

PR #254, opened 2023-04-24
Add crate: `mbedtls-platform-support`

Taowyoo:
To fully enable different `rust-mbedtls` crates to live together.
Need to:

Split logic about exporting rust functions and function pointers in `lib.rs` and `threading.rs` into a separate crate: `mbedtls-platform-support`


mbedtls-platform-support/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+                                                _output: *mut u8) -> i32 {
+    panic!("AES-NI support is forced but the T-tables code was invoked")
+}
+
+
+#[cfg(any(all(feature = "time", feature = "custom_gmtime_r"), sys_time_component = "custom"))]
+#[doc(hidden)]
+#[no_mangle]
+// needs to be pub for global visibility
+pub unsafe extern "C" fn mbedtls_platform_gmtime_r(tt: *const mbedtls_sys::types::time_t, tp: *mut mbedtls_sys::types::tm) -> *mut mbedtls_sys::types::tm {
```
jethrogb, line 63: Is this implementation compatible with 0.7?

jethrogb:
Thoughts on merging this crate with selftest?

Taowyoo:
> Thoughts on merging this crate with selftest?

Yes, it's better to merge them together.

mbedtls-platform-support/src/lib.rs
Taowyoo: Yes.
The function signature actually is same, I just remove the `use` statements.
There is no other import of `time_t` and `tm` 
v0.7 code:
```
#[cfg(all(feature = "time", any(feature = "custom_gmtime_r", feature = "custom_time")))]
use mbedtls_sys::types::time_t;
#[cfg(all(feature = "time", feature = "custom_gmtime_r"))]
use mbedtls_sys::types::tm;

#[cfg(feature="custom_gmtime_r")]
#[doc(hidden)]
#[no_mangle]
pub unsafe extern "C" fn mbedtls_platform_gmtime_r(tt: *const time_t, tp: *mut tm) -> *mut tm {
```

arai-fortanix (approved):
I see that the changes here are moving code from one place to another. So the problems that I noted in the time functions already exist in the code. I'm going to file Jira tickets for the problems I noted, and I don't think they need to be fixed in this change.

Taowyoo:
Going to merge this since, CI on https://github.com/fortanix/rust-mbedtls/pull/260 and https://github.com/fortanix/rust-mbedtls/pull/260 is passed

arai-fortanix (approved):

PR #257, opened 2023-05-04
Fix SPDX license identifiers

DrTobe:
- The [`GPL-2.0+` identifier is deprecated](https://spdx.org/licenses/GPL-2.0+.html) and should be replaced by `GPL-2.0-or-later`
- Using a `/` to separate multiple licenses is not a valid SPDX license expression and is [deprecated for Cargo.toml](https://doc.rust-lang.org/cargo/reference/manifest.html#slash)

We should fix this to enable automatic tools to successfully parse the license field.


Taowyoo (approved):
LGTM
thank you for updating this

Taowyoo:
Fixed by https://github.com/fortanix/rust-mbedtls/pull/263

PR #262, opened 2023-05-09
Make Context Sync

jethrogb:
cc @zvolin

### DIFF
--- a/mbedtls/src/ssl/context.rs
+++ b/mbedtls/src/ssl/context.rs
@@ -126,6 +126,8 @@ pub struct Context<T> {
     client_transport_id: Option<Vec<u8>>,
 }
 
+unsafe impl<T: Sync> Sync for Context<T> {}
+
 impl<'a, T> Into<*const ssl_context> for &'a Context<T> {
     fn into(self) -> *const ssl_context {
         self.handle()
@@ -603,19 +605,19 @@ mod tests {
         assert!(!TestTrait::<dyn Sync, HandshakeContext>::new().impls_trait(), "HandshakeContext must be !Sync");
     }
 
-    struct NonSendStream {
+    struct NonSendSyncStream {
         _buffer: core::ptr::NonNull<u8>,
     }
 
     #[cfg(feature = "std")]
-    impl Read for NonSendStream {
+    impl Read for NonSendSyncStream {
         fn read(&mut self, _: &mut [u8]) -> IoResult<usize> {
             unimplemented!()
         }
     }
     
     #[cfg(feature = "std")]
-    impl Write for NonSendStream {
+    impl Write for NonSendSyncStream {
         fn write(&mut self, _: &[u8]) -> IoResult<usize> {
             unimplemented!()
         }
@@ -625,19 +627,19 @@ mod tests {
         }
     }
 
-    struct SendStream {
+    struct SendSyncStream {
         _buffer: u8,
     }
 
     #[cfg(feature = "std")]
-    impl Read for SendStream {
+    impl Read for SendSyncStream {
         fn read(&mut self, _: &mut [u8]) -> IoResult<usize> {
             unimplemented!()
         }
     }
     
     #[cfg(feature = "std")]
-    impl Write for SendStream {
+    impl Write for SendSyncStream {
         fn write(&mut self, _: &[u8]) -> IoResult<usize> {
             unimplemented!()
         }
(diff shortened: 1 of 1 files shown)


mbedtls/src/ssl/context.rs
```diff
@@ -126,6 +126,8 @@ pub struct Context<T> {
     client_transport_id: Option<Vec<u8>>,
 }
 
+unsafe impl<T: Sync> Sync for Context<T> {}
```
jethrogb, line 129: @Taowyoo let me know if you're comfortable reviewing this change, or we can find someone else

zvolin (approved):
IMO the change is sound given that:
 - `Sync` only targets `&Context` usage, so only methods taking `&self` can be called
 - `Sync` is not implemented for raw pointers more as a 'lint' because those are mostly likely used in FFI, which can do non safe things under the hood

zvolin:
[ssl_get_verify_result](https://github.com/fortanix/rust-mbedtls/blob/e21f0ad0b8ead7c2fbf019a35311e0f05fa6ceba/mbedtls-sys/vendor/library/ssl_tls.c#LL5045C3-L5045C3) - A simple field reading, doesn't mutate state
[ssl_get_alpn_protocol](https://github.com/fortanix/rust-mbedtls/blob/e21f0ad0b8ead7c2fbf019a35311e0f05fa6ceba/mbedtls-sys/vendor/library/ssl_tls.c#L4780) - Same
[ssl_get_bytes_avail](https://github.com/fortanix/rust-mbedtls/blob/e21f0ad0b8ead7c2fbf019a35311e0f05fa6ceba/mbedtls-sys/vendor/library/ssl_msg.c#L5165) - Same

It makes sense that those aren't thread safe in C, because the compiler there can't verify if nothing is mutating the state at the same time, but those should be safe to call in Rust from multiple threads. Please double check if I haven't missed any function that could be used on `&Context` tho :pray:

Taowyoo:
Yes, double checked:  `ssl_get_verify_result`, `ssl_get_alpn_protocol`, `ssl_get_bytes_avail` are the only functions that access mbedtlc C API through `&Context`.
And C code for them is quite simple: only read some simple field

https://github.com/fortanix/rust-mbedtls/blob/81e7f61fa59dfbbc78eede149a25446789c60727/mbedtls-sys/vendor/library/ssl_msg.c#L5165-L5168

https://github.com/fortanix/rust-mbedtls/blob/81e7f61fa59dfbbc78eede149a25446789c60727/mbedtls-sys/vendor/library/ssl_tls.c#L4780-L4783

https://github.com/fortanix/rust-mbedtls/blob/81e7f61fa59dfbbc78eede149a25446789c60727/mbedtls-sys/vendor/library/ssl_tls.c#L5045-L5054

Taowyoo (approved):
Based on @zvolin 's comments, since we ensured that all access `&Context` is thread safe, and thread safety of  access to `&mut Context` will be ensured by rust borrow checker.
It should be OK to add `Sync` impl for `Context<T: Sync>`

Taowyoo:
Note that: user could get a `*const ssl_context` from a `&Context`, it seems we also need to check all generated functions which using `*const ssl_context` are also thread safe

jethrogb:
> it seems we also need to check all generated functions (functions generated by bindgen) which using *const ssl_context are also thread safe

No. Those are all unsafe and caller is required to meet all requirements for safe calling.

zvolin:
I was also thinking about that, but if user would want to do so, then it would involve usage of `unsafe`. I don't think we should bother about that as if someone wants to write unsafe on his own then he has to be careful and verify it's sounds on his own. We cant really guard against that, we can only guard safe api we provide.

However I think it would be good to drop a comment for `rust_mbedtls` developers to double check if their change won't break the rules there

Taowyoo:
Agree too. It worth put a safety comment on top of `unsafe impl<T: Sync> Sync for Context<T> {}`.

mbedtls/src/ssl/context.rs
zvolin, line 129:
```suggestion
/// # Safety
///
/// It is safe for `Context` to be `Sync` as all its methods taking `&self`
/// can only trigger `MbedTLS` functions that have no side effects.
/// When introducing any new FFI call inside `&self` methods, `MbedTLS` sources should
/// be reviewed whether they don't result in any side-effect.
unsafe impl<T: Sync> Sync for Context<T> {}
```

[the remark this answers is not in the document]
Taowyoo: Choose to prefer zvolin's suggestion

Taowyoo:
@jethrogb I am plan to publish the v0.9.1 of `mbedtls` which includes the solution to link conflict problem.
I think this PR is also good to be included in the new version.
Do feel OK to commit the @zvolin 's suggestion and merge this ?
And do you have any other concern?

jethrogb:
If there are no conflicts you don't need to rebase

Yes let's add some comments. Explanation why use of `unsafe` is ok should be a code comment, not a doc comment.

jethrogb:
bors r=YxC

PR #267, opened 2023-05-18
Support decrypting PBES2 key bags in pkcs12

zvolin:
Only supports PBKDF2 key derivation function (without Scrypt).
Supports AES-CBC and DES-EDE3-CBC encryption schemes.

cc @jethrogb


mbedtls/src/pkcs12/mod.rs
```diff
@@ -520,6 +530,26 @@ pub struct Pfx {
...
+enum Pkcs12Pbes {
+    Pbes1 {
+        algo: ObjectIdentifier,
+        params: Pkcs12PbeParams,
+    },
+    Pbes2(Pkcs12Pbes2Params),
+}
+
+impl Pkcs12Pbes {
+    fn decrypt(&self, ciphertext: &[u8], passphrase: &Passphrase) -> Pkcs12Result<Vec<u8>> {
```
zugzwang, line 543: Is it possible to pass a closure computing the passphrase instead of storing it raw in the `Passphrase` object?

mbedtls/src/pkcs12/mod.rs
```diff
@@ -537,75 +567,172 @@ impl BERDecodable for Pkcs12PbeParams {
...
     }
+}
 
-    v
+// See RFC 8018 Appendix A.2
+#[derive(Debug, Clone)]
+struct Pbkdf2Params {
+    salt: Vec<u8>,
+    iterations: u32,
+    #[allow(unused)]
```
zugzwang, line 602: Why `allow(unused)`?

mbedtls/src/pkcs12/mod.rs
```diff
@@ -825,33 +991,49 @@ impl Pfx {
         Ok(decrypted)
     }
 
-    fn authsafe_decrypted_contents(&self) -> impl Iterator<Item=&SafeBag> {
-        self.authsafe.contents.iter()
-            .flat_map(|d| if let AuthenticatedSafe::Data(ref d) = d { d.0.as_slice() } else { &[] })
+    fn authsafe_decrypted_contents(&self) -> impl Iterator<Item = &SafeBag> {
```
zugzwang, line 994: No changes to `fn authsafe_decrypted_contents`, `fn certificates` and `fn private_keys`. Probably due to automatic linting?

mbedtls/src/pkcs12/mod.rs
```diff
@@ -537,75 +567,172 @@ impl BERDecodable for Pkcs12PbeParams {
...
+                encryption_scheme,
+            })
+        })
     }
+}
 
-    v
+// See RFC 8018 Appendix A.2
+#[derive(Debug, Clone)]
+struct Pbkdf2Params {
```
zugzwang, line 599: Regarding parameter validation and regardless of what PBKDF2 does internally; [RFC8018 sec. 4 "Salt and Iteration Count"](https://www.rfc-editor.org/rfc/rfc8018#section-4) provides recommendations for the length of the salt and iteration values (the latter one coming from FIPS 800-132).

Shall we put in place a mechanism to fail early on bad parameters?

jethrogb: This crate is not the right place to implement policy.

mbedtls/src/pkcs12/mod.rs
zvolin: the first commit included is a rustfmt

mbedtls/src/pkcs12/mod.rs
zvolin: while it is possible, there is basically only 1 valid path for the password format for each variant there and `Password` just encapsulates this

zvolin: ah, and also it's not really possible, the closure argument would need to be propagated up to internal `decrypt_data` in `Pfx`, but this function still needs both variants at the same time, to pick the right one depending on which pbe algorithm the `SafeBag` will be encrypted with. We could default to raw password and perform the ucs16 transaltion on the fly where needed. but it's needed in definitely more places than raw one, and each such transform allocs

zvolin: We can also assume that all the passwords passed everywhere are in ucs16 and only transform back to the regular &str form when decoding pbes2. This will remove the `Password` struct and result in only one needless alloc when translating back. Then it'd be worth documenting that all the other places what password format is it expecting, as it was expressed by just &[u8] which doesn't bring this information

zvolin: I could also rename `raw` to `utf8` so that it's clearer

zugzwang: I am not sure that we are talking about the same, but glad for this discussion. (The closure idea was to prompt the end user for a password with caller-defined code, not map to the correct format. Let us drop this idea in any case as out of scope, because this was not before either and perhaps not something to include in rust-mbedtls anyway).

Since we're not going that road, I'm fine with the `Passphrase` struct. The API is easy enough (`Passphrase::new(...)`). I agree with `utf8` instead of `raw`.

Last thing, perhaps it should be called `Pkcs12Passphrase` or similar (in the case where there are other passphrases needed elsewhere in the future).

zvolin: :+1: I'll also rename EncryptionScheme same way

mbedtls/src/pkcs12/mod.rs
```diff
@@ -537,75 +567,172 @@ impl BERDecodable for Pkcs12PbeParams {
...
+        .encode_utf16()
+        .chain(core::iter::once(0))
+        .flat_map(u16::to_be_bytes)
+        .collect()
+}
+
+struct Passphrase<'a> {
+    raw: &'a [u8],
+    ucs_16: Vec<u8>,
+}
```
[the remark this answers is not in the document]
jethrogb: Well, perhaps nevermind. I thought this was for a generic PKCS5 module, but this is all P12. Then we don't have to have any other public API then taking unicode strings.

mbedtls/src/pkcs12/mod.rs
zvolin: done

Taowyoo:
pkcs12 support is removed in `rust-mbedtls`

PR #269, opened 2023-05-19
Remove PKCS12 support

jethrogb:
PKCS#12 is awful. It's a very complicated spec that's not inline with modern cryptographic best practices. It's common for implementations to have unsafe cryptographic defaults, even to this day. RC2, DES, MD5, etc. remain in common use! In addition, common implementations don't follow the password string encoding method from the specification.

There's no reason for us to support this.

This requires a major version bump of the `mbedtls` crate.


Taowyoo (changes requested):
Tests need to be cleaned in `ct.sh` & `mbedtls/tests/pbkdf.rs`

I include this PR in #261 , and you could cherry-pick these two commits to cleanup tests:
- [test(mbedtls): remove pkcs12 related tests](https://github.com/fortanix/rust-mbedtls/pull/213/commits/1f196cbf86614a75595541f224d9f26b2f4d54ef)
- [ci: remove pkcs12 tests](https://github.com/fortanix/rust-mbedtls/pull/213/commits/451d1c8d9ec28fbff8e9e2c7361ddbdd469100ce)

Taowyoo:
Done by https://github.com/fortanix/rust-mbedtls/pull/213

PR #278, opened 2023-06-08
Upgrade `mbedtls` to 3.4.0

Taowyoo:
## Overview

[PR #213](https://github.com/fortanix/rust-mbedtls/pull/213) introduces some breaking changes:
- Update vendor `mbedtls` code from version `2.28.3` to a `3.4.0`
	- Please checkout C `mbedtls` migration guide for 3.X here [3.0-migration-guide.md](https://github.com/fortanix/rust-mbedtls/blob/yx/upgrade-mbedtls/mbedtls-sys/vendor/docs/3.0-migration-guide.md) to
- Update the build code for `mbedtls-sys-auto` crate to sync up with vendor code change:
	- Changes in C DEFINE's for configuration
	- Changes in header files
	- Add binding code generation for `static inline` code in C side
	- Removing deprecated cargo features or dependencies
	- Add a cargo feature for TLS 1.3
- Update rust wrapper code in `./mbedtls` to sync up the changes in  `mbedtls-sys-auto`
	- Removing deprecated cargo features or dependencies
	- Update rust code to sync up API changes
	- Add types and functions for TLS 1.3
	- Add a cargo feature for TLS 1.3
	- Update integration tests for new API and TLS 1.3
	- Update dependencies

## Breaking Changes:

**Note**: entries with :boom: are ensured that they will break the downstream.

#### Vendor code changes

**Upstream changes:**

- :boom: Upgrade vendor `mbeldtls` code to `3.4.0`

**Changes on our side:**

- Cherry picked previous changes in old versions:
	- commit: [vendor change: Adding mpi_force_c_code feature](https://github.com/fortanix/rust-mbedtls/pull/213/commits/c8cd4067ff51f2c74cf623e0a42028215a77f883)
- New changes
    - [vendor change: fix time call in tls13 client&server](https://github.com/fortanix/rust-mbedtls/pull/213/commits/bafc52d97bda9963f36b8eefd19a31110e8e5dac) : This has been merged into upstream, see: https://github.com/Mbed-TLS/mbedtls/pull/7639 .

#### rust code changes

**Features:**

- :boom: `zlib` is removed: support for TLS record-level compression is removed in `mbedtls` 3.X
  - Related C DEFINE `MBEDTLS_ZLIB_SUPPORT` is also removed
- :boom: `legacy_protocols` is removed: all protocols early than TLS 1.2 is removed in `mbedtls` 3.X
  - Related C DEFINE's are also removed: `MBEDTLS_SSL_PROTO_SSL3`, `MBEDTLS_SSL_PROTO_TLS1`, `MBEDTLS_SSL_PROTO_TLS1_1`, `MBEDTLS_SSL_CBC_RECORD_SPLITTING`
- :boom: `pkcs11` is removed: wrapper for `libpkcs11-helper` is removed in `mbedtls` 3.X, see [3.0-migration-guide.md](https://github.com/fortanix/rust-mbedtls/blob/yx/upgrade-mbedtls/mbedtls-sys/vendor/docs/3.0-migration-guide.md#remove-wrapper-for-libpkcs11-helper)
- :boom: ~~`pkcs12` is removed: because https://github.com/fortanix/rust-mbedtls/pull/269~~ `pkcs12` still defined since it is needed for reading some pkcs8 format keys
- Put TLS 1.3 behind a feature `tls13` : because the dependency of TLS 1.3 in mbedtls 3.X are using a global state RNG which breaks the requirements for FIPS, so this feature enables use to avoid these code from compilation.
- :boom: Deprecated features  `custom_threading` , `custom_time` , `custom_gmtime_r` , `pthread`  are removed

**Dependencies:**

- :boom: Bump `mbedtls-sys` version to `3.4.0`
- :boom: Deprecated dependencies are removed
  - `libz-sys` : support for TLS record-level compression is removed in `mbedtls` 3.X
  - `libc`: `libc` is not needed in `sgx`

**Build code changes:**

Following changes are made according to [3.0-migration-guide.md](https://github.com/fortanix/rust-mbedtls/blob/yx/upgrade-mbedtls/mbedtls-sys/vendor/docs/3.0-migration-guide.md).

- Remove `MBEDTLS_CONFIG_H` in `mbedtls_config.h`
- Remove `#include <mbedtls/check_config.h>`
- `mbedtls-sys/build/bindgen.rs`
	-  Allow `bindgen` to generate bindings for functions, types and variables start with `psa_`, and put them in a sub `mod psa` because they are needed by TLS 1.3
	- Use  `bindgen` experiment feature to generate C function wrapper for C `static inline` functions
- `mbedtls-sys/build/headers.rs` : Update header files
- `mbedtls-sys/build/config.rs` : Remove/add C defines
	- :boom: Added `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG` : this is added because TLS 1.3 need to use PSA library which need this when feature `std` is off, user need to provide their implementation. In `rus-mbedtls`, one implementation is proveded, see: `mbedtls/src/rng/mod.rs`.

### Changes in `mbedtls-platform-support`

- Added `once_cell` for initializing PSA only one times: see `fn psa_crypto_init()` in `mbedtls-platform-support/src/lib.rs`, this is needed because:
	- > MBEDTLS_USE_PSA_CRYPTO means that X.509 and TLS will use PSA Crypto as much as possible (that is, everywhere except for features that are not supported by PSA Crypto, see "Internal Changes" below for a complete list of exceptions). When it is enabled, you need to call psa_crypto_init() before calling any function from PK, X.509 or TLS;
	- Ref: https://github.com/Mbed-TLS/mbedtls/blob/0b3de6fcec4aa4b23a9ee1e076714cbc796f3ac4/docs/use-psa-crypto.md#general-considerations
- Add function pointer `mbedtls_psa_external_get_random`  which is needed when C DEFINE option `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG` is turned on. This will be used in case when system default RNG or entropy is no available (for example in SGX)
    - Using function pointer here ensure there is no link time conflict in future when mutiple `rust-mbedtls` is using this crate.
    - User need to call function `set_psa_external_rng_callback` before using any PSA functions or TLS 1.3
- Add rust implementation of `explicit_bzero`, which is needed in SGX. Because in SGX, our [rs-libc](https://github.com/fortanix/rust-sgx/tree/master/rs-libc) does not support this function.
- Update self tests:
  - following are removed:
    - arc4_self_test
    - md2_self_test
    - md4_self_test
    - x509_self_test
    - xtea_self_test
  - following are added:
    - sha384_self_test
    - chacha20_self_test
    - chachapoly_self_test
    - poly1305_self_test
    - sha224_self_test

### Changes in `mbedtls`

**Features:**

- :boom: `zlib`, `legacy_protocols` removed to sync with changes in `mbedtls-sys`
- :boom: `pkcs12` and `pkcs12_rc2` are removed: see https://github.com/fortanix/rust-mbedtls/pull/269
- Put TLS 1.3 logic behind a feature `tls13`: check reason above

**Dependencies:**

- :boom: Bump `rust-mbedtls` version to `0.11.0`
- :boom: Bump dependency `mbedtls-sys-auto` version to `3.4.0`
- Added `rstest` `rstest_reuse` `lazy_static` `env_logger`: used for improving code of tests

### Code changes

- Function `EcPoint::mul` now need to pass in a RNG to ensure blinding.
- Add prefix `private_` to some fields of `mbedtls` types under `mbedtls/src/cipher/raw`
- :boom: Removed and added some options in `Error`, `CipherType`, `CipherMode` and `CipherSuite`  to sync with changes in `mbedtls` 3.X
- :boom: `mbedtls/src/pk/ec.rs` : User need to provide a RNG callback for function `EcPoint::mul`, this originally is not a hard requirement, but in C `mbedtls` 3.X this become a hard requirement for blinding to defend side channel attack.
- :boom: `mbedtls/hash` :
	- `Md2` and `Md4` are removed since they are no longer supported in `mbedtls` 3.X
	- fn `pbkdf_pkcs12` is removed since `pkcs12` is removed
- :boom: `mbedtls/pk/mod.rs` :
	- Remove `CustomPk`
	- User need to provide a RNG callback for `Pk::from_private_key`, this originally is not a hard requirement, but in C `mbedtls` 3.X this become a hard requirement for blinding to defend side channel attack.
- :boom: `mbedtls/src/ssl/ciphersuites.rs`: Rename `TlsCipherSuite` to `Tls12CipherSuite`, and add enum: `Tls13CipherSuite`, `IanaTlsNamedGroup`, `TLS13SignatureAlgorithms`: these are introduced by TLS 1.3
- `mbedtls/src/ssl/ssl_states.rs`: Add `SslStates` to represent the state of SSL handshake
- Update tests accordingly

**Special code need to notice**:

In `impl` of `std::io::Read` under `mbedtls/src/ssl/io.rs` and `tokio::io::AsyncRead` under `mbedtls/src/ssl/async_io.rs`, there are some code to handle the special case when using `mbedtls` as `client` to connect to a server whose `session ticket` extension is enabled.

This case is found when connecting to `goolge.com`, where Google's server send the `session ticket` after the completion of handshake, which cause `C-mbedtls` throw errors when client is try to read msg data.

## CI changes

- Use [cargo-nextest](https://nexte.st/#cargo-nextest) to run tests
	- Reduce time to run tests
	- Have ability to run some tests in serial
		- tests under `hyper.rs` need to access to `google.com` which has QPS limit
		- some tests function use some system resource, see https://github.com/fortanix/rust-mbedtls/blob/yx/upgrade-mbedtls/mbedtls/tests/support/net.rs


mbedtls-platform-support/src/psa_external_rng.rs
```diff
@@ -0,0 +1,45 @@
...
+type PsaExternlRngCallback = unsafe extern "C" fn(
+    user_data: *mut c_void,
+    data: *mut c_uchar,
+    len: size_t,
+    olen: *mut size_t,
+) -> int32_t;
+
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub static mut mbedtls_psa_external_get_random: PsaExternlRngCallback = PsaExternlRng::dummy_callback;
```
mzohreva, line 36: Why `pub`?

mbedtls-platform-support/src/psa_external_rng.rs
```diff
@@ -0,0 +1,45 @@
...
+ * 2.0 <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0>, at your
+ * option. This file may not be copied, modified, or distributed except
+ * according to those terms. */
+
+use mbedtls_sys::types::raw_types::{c_void, c_uchar};
+use mbedtls_sys::types::{size_t, int32_t};
+use once_cell::sync::OnceCell;
+static PSA_CALLBACK_INIT: OnceCell<()> = OnceCell::new();
+
+struct PsaExternlRng {}
```
mzohreva, line 14: What's the purpose of this type? it doesn't seem to be used

mbedtls-sys/build/bindgen.rs
```diff
@@ -20,7 +19,21 @@ struct MbedtlsParseCallbacks;
 
 impl bindgen::callbacks::ParseCallbacks for MbedtlsParseCallbacks {
     fn item_name(&self, original_item_name: &str) -> Option<String> {
-        Some(original_item_name.trim_start_matches("mbedtls_").trim_start_matches("MBEDTLS_").to_owned())
+        // clean up generated static function suffix
+        let original_item_name = original_item_name.trim_end_matches("__extern");
```
mzohreva, line 23: this is not properly named, maybe: `let item_name = ...`

mbedtls-sys/build/bindgen.rs
```diff
@@ -20,7 +19,21 @@ struct MbedtlsParseCallbacks;
...
+        // remove reductant prefix for code in mbedtls lib
```
mzohreva, line 24: typo: reductant -> redundant

mbedtls-sys/build/bindgen.rs
```diff
@@ -20,7 +19,21 @@ struct MbedtlsParseCallbacks;
...
+            original_item_name.trim_start_matches("mbedtls_").to_string()
+        } else if original_item_name.starts_with("MBEDTLS_") {
+            original_item_name.trim_start_matches("MBEDTLS_").to_string()
+        // remove reductant prefix for code in psa lib
+        } else if original_item_name.starts_with("psa_") {
+            original_item_name.trim_start_matches("psa_").to_string()
+        } else if original_item_name.starts_with("PSA_") {
+            original_item_name.trim_start_matches("PSA_").to_string()
+        } else {
+            original_item_name.to_string()
```
mzohreva, line 35: the if statements are not necessary, you can do this:
```
original_item_name.trim_end_matches("__extern")
    .trim_start_matches("mbedtls_")
    .trim_start_matches("MBEDTLS_")
    .trim_start_matches("psa_")
    .trim_start_matches("PSA_")
    .to_owned()
```

Taowyoo: They are not identical. For example in case a function called `psa_mbedtls_foo` it will be changed `foo`.

The if statement ensure the trim only run once on each input

mbedtls-platform-support/src/psa_external_rng.rs
Taowyoo: I see what you mean, I could just use a normal function for this.
will remove this.

mbedtls/src/bignum/mod.rs
```diff
@@ -133,22 +133,23 @@ impl Mpi {
         Ok(())
     }
 
-    fn get_limb(&self, n: usize) -> mpi_uint {
-        if n < self.inner.n {
-            unsafe { *self.inner.p.offset(n as isize) }
```
mzohreva, line 138: is this not available anymore?

mbedtls/src/bignum/mod.rs
```diff
@@ -133,22 +133,23 @@ impl Mpi {
...
-
     pub fn as_u32(&self) -> Result<u32> {
         if self.bit_length()? > 32 {
             // Not exactly correct but close enough
             return Err(codes::MpiBufferTooSmall.into());
         }
 
-        Ok(self.get_limb(0) as u32)
+        let mut buf = [0u8; 4];
+        if cfg!(target_endian = "little")
```
mzohreva, line 143: Why do you need this check? can't you use either `mpi_write_binary_le/u32::from_le_bytes` or `mpi_write_binary/u32::from_be_bytes` regardless of target?

mbedtls/src/cipher/mod.rs
```diff
@@ -290,12 +290,12 @@ impl Cipher<Encryption, Authenticated, AdditionalData> {
     pub fn encrypt_auth_inplace(
         mut self,
         ad: &[u8],
-        data: &mut [u8],
-        tag: &mut [u8],
+        data_with_tag: &mut [u8],
```
mzohreva, line 293: This merits a doc comment explaining the order of data and tag in the input slice

egrimley-arm:
With some workflows it can be inconvenient when a non-empty source file does not finish with a newline character. For example, an editor might add the newline character and cause a spurious diff. So it would be nice to add a final newline character to these files:
```
.config/nextest.toml
3.X-migration-guide.md
Cargo.toml
ct_locally.sh
mbedtls-platform-support/src/psa_external_rng.rs
mbedtls/src/error.rs
mbedtls/src/rng/mod.rs
mbedtls/src/ssl/mod.rs
mbedtls/tests/support/debug.rs
```

egrimley-arm:
This symbolic link was presumably added by accident:
```
mbedtls-sys/vendor/build/library/ssl_debug_helpers_generated.c
```

mbedtls-sys/build/config.rs
```diff
@@ -421,25 +399,29 @@ pub fn default_defines() -> HashMap<&'static str, Macro> {
...
     ("trusted_cert_callback", ("MBEDTLS_X509_TRUSTED_CERTIFICATE_CALLBACK", Defined)),
+    ("tls13",                 ("MBEDTLS_PSA_CRYPTO_C",                                       Defined)),
+    ("tls13",                 ("MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG",                            Defined)),
+    ("tls13",                 ("MBEDTLS_SSL_PROTO_TLS1_3",                                   Defined)),
+    ("tls13",                 ("MBEDTLS_SSL_TLS1_3_COMPATIBILITY_MODE",                      Defined)),
+    ("tls13",                 ("MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_ENABLED",           Defined)),
+    ("tls13",                 ("MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_EPHEMERAL_ENABLED",     Defined)),
+    ("tls13",                 ("MBEDTLS_SSL_TLS1_3_KEY_EXCHANGE_MODE_PSK_EPHEMERAL_ENABLED", Defined)),
+    ("tls13",                 ("MBEDTLS_SSL_TLS1_3_TICKET_AGE_TOLERANCE",                    DefinedAs("6000"))),
+    ("tls13",                 ("MBEDTLS_SSL_TLS1_3_TICKET_AGE_TOLERANCE",                    DefinedAs("6000"))),
```
mzohreva, line 422: duplicate line

mbedtls-platform-support/src/lib.rs
```diff
@@ -106,3 +106,28 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut mbedtls_sys::types::time_t) -> mb
...
+// needs to be pub for global visibility
+pub unsafe extern "C" fn explicit_bzero(buf: *mut mbedtls_sys::types::raw_types::c_void, len: mbedtls_sys::types::size_t)  {
+    use zeroize::Zeroize;
+    let buffer = core::slice::from_raw_parts_mut(buf as *mut mbedtls_sys::types::raw_types::c_char, len);
+    buffer.zeroize();
+}
+
+/// you need to call `psa_crypto_init()` before calling any function from the SSL/TLS, X.509 or PK modules
+#[cfg(feature = "tls13")]
+pub fn psa_crypto_init() {
```
mzohreva, line 122: re-export this from mbedtls crate for convenience

mbedtls-platform-support/src/lib.rs
```diff
@@ -106,3 +106,28 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut mbedtls_sys::types::time_t) -> mb
...
+#[doc(hidden)]
+#[no_mangle]
+// needs to be pub for global visibility
+pub unsafe extern "C" fn explicit_bzero(buf: *mut mbedtls_sys::types::raw_types::c_void, len: mbedtls_sys::types::size_t)  {
+    use zeroize::Zeroize;
+    let buffer = core::slice::from_raw_parts_mut(buf as *mut mbedtls_sys::types::raw_types::c_char, len);
+    buffer.zeroize();
+}
+
+/// you need to call `psa_crypto_init()` before calling any function from the SSL/TLS, X.509 or PK modules
```
mzohreva, line 120: add comment that it's ok to call this multiple times

mbedtls-platform-support/src/lib.rs
```diff
@@ -106,3 +106,28 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut mbedtls_sys::types::time_t) -> mb
...
+    use once_cell::sync::OnceCell;
+    static INIT: OnceCell<()> = OnceCell::new();
+
+    INIT.get_or_init(|| {
+        unsafe { mbedtls_sys::psa::crypto_init() };
+        return ();
+    });
+}
+
+#[cfg(all(any(feature = "rdrand", target_env = "sgx"), feature = "tls13"))]
```
mzohreva, line 132: clarify and document why exactly we enable this under these conditions

mbedtls-platform-support/src/self_test.rs
```diff
@@ -94,12 +94,32 @@ pub unsafe fn disable() {
...
+    chachapoly_self_test as chachapoly,
+    ecjpake_self_test as ecjpake,
+    ctr_drbg_self_test as ctr_drbg,
+    hmac_drbg_self_test as hmac_drbg,
+    pkcs5_self_test as pkcs5,
+    nist_kw_self_test as nist_kw,
+    des_self_test as des,
+    camellia_self_test as camellia,
+    base64_self_test as base64,
+    aria_self_test as aria,
```
mzohreva, line 124: please order by name

mzohreva:
note to self: reviewed platform and sys changes

mbedtls/src/cipher/raw/mod.rs
```diff
@@ -168,13 +205,26 @@ define!(
 define!(
     #[c_ty(cipher_context_t)]
     #[repr(C)]
-    struct Cipher;
-    const init: fn() -> Self = cipher_init;
+    struct Cipher {
+        iv: Option<Vec<u8>>,
```
zugzwang, line 209: Why is it now necessary to store the IV on the rust object? It is normally stored inside the mbedtls object after `set_key_iv`

zugzwang: e.g. it may also be present in `inner`

mbedtls/src/cipher/raw/mod.rs
```diff
@@ -276,28 +327,27 @@ impl Cipher {
...
-        unsafe {
-            if (*self.inner.cipher_info).mode == MODE_GCM
-                || (*self.inner.cipher_info).mode == MODE_CCM
-            {
-                return true;
-            } else {
-                return false;
-            }
+        match self.cipher_mode() {
+            CipherMode::GCM | CipherMode::CCM => true,
```
zugzwang, line 349: ChaChaPoly is also an authenticated mode.

mbedtls/src/ecp/mod.rs
```diff
@@ -335,8 +334,8 @@ impl EcPoint {
                 &mut ret.inner,
                 k.handle(),
                 &self.inner,
-                None,
-                ::core::ptr::null_mut(),
+                Some(F::call),
```
zugzwang, line 337: this change is important security-wise; as we were not blinding before... perhaps highlight it in the changelog.

mbedtls/src/lib.rs
```diff
@@ -79,6 +78,13 @@ mod mbedtls {
     pub use super::*;
 }
 
+/// You need to call `psa_crypto_init()` before calling any function from the SSL/TLS, X.509 or PK modules.
+/// This function is fine to be called mutiple times while ensure underlying initilization function is only
+/// been called only once.
+/// Re-export here for convenient.
```
zugzwang, line 84: nit: convenient -> convenience

mbedtls/src/pk/mod.rs
```diff
@@ -236,7 +185,7 @@ impl Pk {
     /// Takes both DER and PEM forms of PKCS#1 or PKCS#8 encoded keys.
     ///
     /// When calling on PEM-encoded data, `key` must be NULL-terminated
-    pub fn from_private_key(key: &[u8], password: Option<&[u8]>) -> Result<Pk> {
+    pub fn from_private_key<F: Random>(rng: &mut F, key: &[u8], password: Option<&[u8]>) -> Result<Pk> {
```
zugzwang, line 188: This one is odd from the user perspective. I have some key bytes and in order to read them, I need to provide an RNG.

Randomness here is actually needed for computing the public key given the read private key bytes (for instance, in EC, private key is a scalar `d` and public key is `dP`, the scalar multiplication of  `d`. To protect it against side channels, blinding is used when multiplying, which requires randomness).

Ideally, IMHO, this function should only read the private bytes (no randomness), and compute/cache the public key (with the RNG) when a function is called to compute the public key for a given `Pk` (e.g. `.public_key`. I don't know if this is currently possible, given MbedTLS's plans about a global RNG. We should check the internals of `pk_parse_key` and see if there are similar funcions that only read the private key.

mbedtls/src/pk/mod.rs
```diff
@@ -295,16 +246,18 @@ impl Pk {
         Ok(ret)
     }
 
-    pub fn private_from_ec_components(mut curve: EcGroup, private_key: Mpi) -> Result<Pk> {
+    // This function now need a RNG because it utilized `EcPoint::mul` function which must need
+    // a random number generator for blinding in mbedtls 3.X, 
+    pub fn private_from_ec_components<F: Random>(rng: &mut F, mut curve: EcGroup, private_key: Mpi) -> Result<Pk> {
```
zugzwang, line 251: Same applies here. `rng` is only needed when computing public data from private bytes.

mbedtls/src/lib.rs
```diff
@@ -7,6 +7,8 @@
  * according to those terms. */
 
 #![deny(warnings)]
+// to be removed after migration to rustc 1.70.., issue #277
+#![allow(ambiguous_glob_reexports)]
```
egrimley-arm, line 11: The combination of `#![deny(warnings)]` and `#![allow(ambiguous_glob_reexports)]` gives an error with older versions of Rust:
```
error: unknown lint: `ambiguous_glob_reexports`
```
However, with `#![allow(unknown_lints)]` inserted between those two attributes it seems to work nicely with Rust 1.65, at least, and perhaps it would for everyone that way, so consider adding that line.

Taowyoo:
About comments about newly add `RNG` arguments for `mul` `from_private_key` `private_from_ec_components` , I think:
in `mbedtls` , the changes are fine because they just represent the API change from upstream and current PR will be a major version bump, so it's OK to have such change. 
Some better comments about these public API are added to explain this clearly.

Taowyoo:
bors cancel

Taowyoo:
bors retry

Taowyoo:
bors try

Taowyoo:
bors cancel

Taowyoo:
bors cancel

Taowyoo:
bors try-

Taowyoo:
bors try-

Taowyoo:
bors cancel

Taowyoo:
bors try

mbedtls/src/ssl/async_io.rs
```diff
@@ -95,17 +95,43 @@ where
...
         }
 
-        self.with_bio_async(cx, |ssl_ctx| match ssl_ctx.recv(buf.initialize_unfilled()) {
-            Err(e) if e.high_level() == Some(codes::SslPeerCloseNotify) => Poll::Ready(Ok(())),
-            Err(e) if e.high_level() == Some(codes::SslWantRead) => Poll::Pending,
-            Err(e) => Poll::Ready(Err(crate::private::error_to_io_error(e))),
-            Ok(i) => {
-                buf.advance(i);
-                Poll::Ready(Ok(()))
+        self.with_bio_async(cx, |ssl_ctx| loop {
```
raoulstrackx, line 102: Nitpick: I proposed a recursive call for the `read` function in `io.rs`. I'd take the same route here for consistency

mbedtls-platform-support/src/psa_external_rng.rs
```diff
@@ -0,0 +1,40 @@
...
+    data: *mut c_uchar,
+    len: size_t,
+    olen: *mut size_t,
+) -> int32_t;
+
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub static mut mbedtls_psa_external_get_random: PsaExternlRngCallback = dummy_callback;
+
+pub fn set_psa_external_rng_callback(
```
jethrogb, line 34: This is not great

mbedtls-platform-support/src/lib.rs
```diff
@@ -106,3 +106,30 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut mbedtls_sys::types::time_t) -> mb
     }
     timestamp
 }
+
+#[cfg(target_env = "sgx")]
+#[doc(hidden)]
+#[no_mangle]
+// needs to be pub for global visibility
+pub unsafe extern "C" fn explicit_bzero(buf: *mut mbedtls_sys::types::raw_types::c_void, len: mbedtls_sys::types::size_t)  {
```
jethrogb, line 114: Shouldn't this be in rs-libc?

Taowyoo:
Added back `pkcs12` C define because this is needed for importing `RSA 2048 Pkcs8 encrypted key with pbeWithSHAAnd3-KeyTripleDES-CBC`, see test:
https://github.com/fortanix/rust-mbedtls/blob/8538c469d7679071840fc460aa56eb5bb5759d72/mbedtls/src/pk/mod.rs#L1197-L1231

mbedtls-sys/build/config.rs
```diff
@@ -421,25 +399,28 @@ pub fn default_defines() -> HashMap<&'static str, Macro> {
...
-    ("legacy_protocols",      ("MBEDTLS_SSL_PROTO_SSL3",                    Defined)),
-    ("legacy_protocols",      ("MBEDTLS_SSL_PROTO_TLS1",                    Defined)),
-    ("legacy_protocols",      ("MBEDTLS_SSL_PROTO_TLS1_1",                  Defined)),
-    ("legacy_protocols",      ("MBEDTLS_SSL_CBC_RECORD_SPLITTING",          Defined)),
     ("aes_alt",               ("MBEDTLS_AES_ENCRYPT_ALT",                   Defined)),
     ("aes_alt",               ("MBEDTLS_AES_DECRYPT_ALT",                   Defined)),
     ("mpi_force_c_code",      ("MBEDTLS_MPI_FORCE_C_CODE",                  Defined)),
     ("trusted_cert_callback", ("MBEDTLS_X509_TRUSTED_CERTIFICATE_CALLBACK", Defined)),
+    ("tls13",                 ("MBEDTLS_PSA_CRYPTO_C",                                       Defined)),
+    ("tls13",                 ("MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG",                            Defined)),
```
jethrogb, line 415: I think this should only be enabled on platforms that don't have platform entropy (e.g. SGX)

jethrogb: Or perhaps the `undefined` below already takes care of that?

mbedtls-platform-support/src/lib.rs
```diff
@@ -106,3 +106,30 @@ pub unsafe extern "C" fn mbedtls_time(tp: *mut mbedtls_sys::types::time_t) -> mb
...
+    use once_cell::sync::OnceCell;
+    static INIT: OnceCell<()> = OnceCell::new();
+
+    INIT.get_or_init(|| {
+        unsafe { mbedtls_sys::psa::crypto_init() };
+        return ();
+    });
+}
+
+#[cfg(all(any(feature = "rdrand", target_env = "sgx"), feature = "tls13"))]
```
jethrogb, line 134: This should only depend on whether mbedtls-sys was built with `MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG`. You can use features passed up through the build scripts.

mbedtls-platform-support/src/rng/mod.rs
```diff
@@ -0,0 +1,22 @@
...
+
+#[cfg(all(any(feature = "rdrand", target_env = "sgx"), feature = "tls13"))]
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub unsafe extern "C" fn mbedtls_psa_external_get_random(
+    _user_data: *mut mbedtls_sys::types::raw_types::c_void,
+    data: *mut mbedtls_sys::types::raw_types::c_uchar,
+    len: mbedtls_sys::types::size_t,
+    _olen: * mut mbedtls_sys::types::size_t) -> mbedtls_sys::types::int32_t {
+    *_olen = len;
```
jethrogb, line 20: I don't think you should set this unconditionally (there may be an error)

mbedtls/src/lib.rs
```diff
@@ -79,6 +82,13 @@ mod mbedtls {
     pub use super::*;
 }
 
+/// You need to call `psa_crypto_init()` before calling any function from the SSL/TLS, X.509 or PK modules.
+/// This function is fine to be called mutiple times while ensure underlying initilization function is only
+/// been called only once.
```
jethrogb, line 87: Is that so? I thought this was only needed for TLS, where we already call it?

We can't put this kind of requirement on our users. If this is needed, we need to call it ourselves.

mbedtls/tests/support/debug.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// Mbed TLS has a feature to show the TLS handshake logs, filtering with
+/// certain debug level. Note that `DEFAULT_MBEDTLS_DEBUG_LEVEL` is the default
+/// level of debug logs if you did not specify it through `MBEDTLS_DEBUG`
+/// environment variable.
+#[cfg(feature = "debug")]
+pub const DEFAULT_MBEDTLS_DEBUG_LEVEL: i32 = 3;
+
+#[cfg(feature = "debug")]
+lazy_static::lazy_static! {
```
TylerBloom, line 20: Minor nit: should be replaced with a `OnceCell` or `Oncecell::Lazy`

mbedtls/src/lib.rs
```diff
@@ -7,6 +7,12 @@
  * according to those terms. */
 
 #![deny(warnings)]
+// TODO: this is added to ensure `#![allow(ambiguous_glob_reexports)]` also works
+// in rust with old version, need to be removed after migration to rustc 1.70..,
+// See issue #277
+#![allow(unknown_lints)]
```
TylerBloom, line 13: After stabilization (i.e. during the 3.5.0), more `deny` directives should be added. IMO, a decent starting list is:
```rust
#![deny(
    unused,
    dead_code,
    missing_docs,
    missing_debug_implementations,
    rustdoc::broken_intra_doc_links,
    unreachable_pub,
    unused_allocation,
    unused_comparisons,
    unused_parens,
    while_true,
    trivial_casts,
    trivial_numeric_casts,
    unused_extern_crates,
    unused_import_braces,
    unused_qualifications,
    unused_results,
    clippy::all,
)]
#![warn(rust_2018_idioms)]
```

TylerBloom (commented):
I reviewed the changes to the `mbedtls` subcrate. I can't speak to correctness/soundness of the crypto logic, but the Rust logic seems sound. The only comments that I have are relatively minor or are suggestions for more general improvements, which are probably outside of the scope of this PR (i.e. using Miri in CI and adding `deny` directives).

Overall, the `mbedtls` code changes look good to me.

TylerBloom (approved):

Taowyoo:
bors cancel

Taowyoo:
bors merge

PR #279, opened 2023-06-09
Add accessors in vendor code to replace `private_` fields access

Taowyoo:
According to https://github.com/fortanix/rust-mbedtls/blob/f92979a91102fd3e743decee1c0ecfd55b3d29dd/mbedtls-sys/vendor/docs/3.0-migration-guide.md

> ### Most structure fields are now private
> 
> Direct access to fields of structures (`struct` types) declared in public headers is no longer supported. In Mbed TLS 3, the layout of structures is not considered part of the stable API, and minor versions (3.1, 3.2, etc.) may add, remove, rename, reorder or change the type of structure fields.
> 
> There is a small number of exceptions where some fields are guaranteed to remain stable throughout the lifetime of Mbed TLS 3.x. These fields are explicitly documented as public. Please note that even if all the fields of a structure are public, future versions may add new fields. Also, as before, some public fields should be considered read-only, since modifying them may make the structure inconsistent; check the documentation in each case.
> 
> Attempting to access a private field directly will result in a compilation error.
> 
> If you were accessing structure fields directly, and these fields are not documented as public, you need to change your code. If an accessor (getter/setter) function exists, use that. Direct accessor functions are usually called `mbedtls_<MODULE>_{get,set}_<FIELD>` or `mbedtls_<MODULE>_<STRUCTURE>_{get,set}_<FIELD>`. Accessor functions that change the format may use different verbs, for example `read`/`write` for functions that import/export data from/to a text or byte string.
> 
> If no accessor function exists, please open an [enhancement request against Mbed TLS](https://github.com/Mbed-TLS/mbedtls/issues/new?template=feature_request.md) and describe your use case. The Mbed TLS development team is aware that some useful accessor functions are missing in the 3.0 release, and we expect to add them to the first minor release(s) (3.1, etc.).
> 
> As a last resort, you can access the field `foo` of a structure `bar` by writing `bar.MBEDTLS_PRIVATE(foo)`. Note that you do so at your own risk, since such code is likely to break in a future minor version of Mbed TLS.

We need to prevent accessing to any field with prefix: `private_`, so in this PR, a lot of accessors have been added to vendor code to achieve this.

All vendor code also need to be up-streamed


Taowyoo:
This PR is too old, will create a another new PR for https://github.com/fortanix/rust-mbedtls/issues/283

PR #292, opened 2023-07-10
Use fips to replace tls13 feature

Taowyoo:
Instead of using `tls13` to explicitly enable TLS 1.3. It's better to use `fips` to specify the limitation on TLS 1.3 since currently TLS 1.3 implementation in C `mbedtls` breaks the requirement of FIPS.


jethrogb:
I don't understand the rationale

Taowyoo:
Hi @jethrogb , I mean use `fips` for conditionally compile TLS 1.3 code is a better approach. 
In current cargo ,for user crate, there is no way to turn off a feature explicitly (in our use case).

PR #294, opened 2023-07-11
Set `MBEDTLS_SSL_OUT_CONTENT_LEN` to a smaller value to avoid client side record overflow

### LINKED ISSUE #293: Bug: `mbedlts` sometimes breaks the record size limit
In TLS 1.3, common TLS implementations now send `record_size_limit` extension to negotiation max record size between client and server. But:

- `mbedtls` currently (ver: 3.4.0) does not support it and just ignore that extension.
- also sometimes `mbedlts` will send a TLS record with **record length** > **2^14** which breaks the RFC, as a result, causing  client's record overflow error. 

This causes problem: clients (especially common browsers) will throw "record_overflow" alert to server when server try to send a record with too big record size.

So current temporary solution is to manually set the `MBEDTLS_SSL_OUT_CONTENT_LEN` to a smaller number, such as `4096` ( this number comes from https://www.openssl.org/docs/man1.1.1/man3/BIO_set_write_buffer_size.html )

This may also related to the our `async_write` function under `mbedtls/src/ssl/context.rs`.

Ref:
https://www.rfc-editor.org/rfc/rfc8446#section-5.1
https://datatracker.ietf.org/doc/html/rfc8449

Taowyoo:
In some cases, mbedlts will send a tls record with record length > 2^14 which breaks the rfc, as a result trigger client's record overflow error. This PR is a temporary fix for it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -641,7 +641,7 @@ dependencies = [
 
 [[package]]
 name = "mbedtls-sys-auto"
-version = "3.5.0-alpha.2+0b3de6f"
+version = "3.5.0-alpha.3+0b3de6f"
 dependencies = [
  "bindgen",
  "cc",
--- a/mbedtls-sys/Cargo.toml
+++ b/mbedtls-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mbedtls-sys-auto"
-version = "3.5.0-alpha.2+0b3de6f"
+version = "3.5.0-alpha.3+0b3de6f"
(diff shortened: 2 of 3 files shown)


arai-fortanix (commented):
Can you please explain how this fixes the issue? Does mbedtls already support `MBEDTLS_SSL_OUT_OUT_CONTENT_LEN` if it's set?

Taowyoo:
Hi @arai-fortanix , yes, it's provide by `mbedtls`, originally we did not set it and it was configured to default value of 2^14

arai-fortanix (approved):

PR #296, opened 2023-07-12
Improve CI

Taowyoo:
- Add `cargo-cache` for `test` workflow.
- Correct and improve publish crate workflow.
- Improve nextest config

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -19,7 +19,7 @@ serial-integration = { max-threads = 1 }
 # * retries = 3
 # * retries = { backoff = "fixed", count = 2, delay = "1s" }
 # * retries = { backoff = "exponential", count = 10, delay = "1s", jitter = true, max-delay = "10s" }
-retries = 2
+retries = 0
 
 # The number of threads to run tests with. Supported values are either an integer or
 # the string "num-cpus". Can be overridden through the `--test-threads` option.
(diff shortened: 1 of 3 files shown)


.config/nextest.toml
```diff
@@ -111,8 +111,10 @@ store-success-output = false
 store-failure-output = true
 
 [[profile.default.overrides]]
-filter = 'binary(=hyper)'
-retries = { backoff = "fixed", count = 6, delay = "1s" }
+# `hyper` test targeting external Google's service, it would sometomes failed because of rate limit
```
arai-fortanix, line 114: Unit tests really shouldn't rely on external services, especially ones that might rate limit us. Can this test be changed so it runs its own server locally that the test can make requests to? Ideally, the server should use a different network implementation, so something like python.

.github/workflows/publish-crates.yml
```diff
@@ -13,24 +17,42 @@ jobs:
...
-      - uses: actions-rs/toolchain@v1
-        with:
-            toolchain: stable
-            override: true
-      - run: |
-          export crate_name=$(python3 -c "print('$GITHUB_REF'.split('/')[2].rsplit('_v', 1)[0])")
-          echo "Publishing crate: $crate_name"
-          cargo publish --locked --token ${CARGO_REGISTRY_TOKEN} --package "$crate_name"
+          # Create a symbolic link to ensure correct usage of libclang
+          sudo ln -s /lib/x86_64-linux-gnu/libclang-11.so.1 /lib/x86_64-linux-gnu/libclang.so
```
arai-fortanix, line 35: Since the .so needs to be there, just possibly not in the location where it's being looked for, can this be fixed by adding the appropriate directory to the search path for the linker and/or setting LD_LIBRARY_PATH when whatever needs this library is being run? Adding symlinks to standard system locations seems like not a good idea for a build/test system to be doing.

arai-fortanix (approved):

.config/nextest.toml
Taowyoo: Make sense.
Created an issue for this: https://github.com/fortanix/rust-mbedtls/issues/297

.github/workflows/publish-crates.yml
Taowyoo: I see, after recheck. These lines seems not necessary.

PR #300, opened 2023-07-18
Fix builds and add CI check for no-std targets

DrTobe:
The current `master` does not build for `no_std` targets because `mbedtls-platform-supports` uses [`once_cell::sync::OnceCell`](https://github.com/fortanix/rust-mbedtls/blob/5c18defba8032aea3e6b93dea9d44d2afc325e1c/mbedtls-platform-support/src/lib.rs#L126), the `once_cell::sync` module ist only available if the `std` feature is active for `once_cell`. Thus, the current `master` only builds for `std` targets.

This PR fixes this, but only if the `tls13` feature is not active. (Edit: Not true anymore thanks to the `critical-section` feature in `once_cell`)


Taowyoo:
The PR LGTM.
Oh, i see `no_std` not equals `no_std_deps`.
Could you also add the CI for this case?

DrTobe:
> If possible, need to add a build check on this in CI to avoid same problem in future

That sounds like a great idea! I will try to add a corresponding build target in the CI when I find some time to work on that.

DrTobe:
Fixing no_std support apparently requires more than I first anticipated. The current state is not safe to merge because many types `impl Sync` under the false assumption that `mbedtls-sys` is used with the "threading" feature (thus I converted this into "Draft" state). But before I can fix this I need to understand which parts needs to be made dependent on the "threading" feature and if more changes are necessary. This requires deeper understanding of the mbedtls codebase and unraveling some of the macro magic involved. In short, this will take some more time to be finished.

xiaoyuxlu:
@DrTobe @Taowyoo 
I also have a problem with ```no_std``` build as well. #302 
I suggest fixing ```no_std``` build first #303 . This will unblock the use of ```no_std```. 
The next step is to enable CI and ```tls13``` in no_std. what do you think?

Taowyoo:
Close this old PR because of change of default branch to `main`.
See more:
- https://github.com/fortanix/rust-mbedtls/issues/320

Please reopen this if have further problems or you rebased this PR onto `main`(this is needed on `mbedtls 2.28.X`)

PR #308, opened 2023-08-21
fix: return error when verify empty cert chain

### LINKED ISSUE #307: `mbedtls::x509::certificate::Certificate::verify()` segfaults when passing in an empty certificate chain to verify
The first argument to the `x509_crt_verify()` call is `chain.inner_ffi_mut()`. If this chain is an empty `MbedtlsList` (e.g., one created using `MbedtlsList::new()`), then the `inner_ffi_mut()` call will return a null pointer. This null pointer will be dereferenced inside the `x509_crt_verify()` call, causing a segmentation fault.

Although the documentation of `mbedtls_x509_crt_verify()` doesn’t state it explicitly, it looks like passing a null pointer for its first argument is not allowed. Hence this appears to be a `rust-mbedtls` bug, not an `mbedtls` bug.

Proposed solution: modify the `verify_ex()` function in `rust-mbedtls` to return a `X509BadInputData` error if the input chain is empty.

The fix PR need to back-port to 0.9 and 0.8:
- [x] master
- [x] v0.9
- [x] v0.8

Taowyoo:
For #307 on master.

Several back-port PRs needed for older versions.

Only return X509BadInputData error when candidate certificate chain is empty because:
- underlying `mbedtls` does not have null pointer check on it.
- underlying `mbedtls` has null pointer check on `trust_ca` chain during the process of finding parent certificate in the chain.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test
+name: CI
 
 on:
   push:
@@ -19,8 +19,8 @@ env:
   CARGO_NET_RETRY: 10
 
 jobs:
-  build:
-    name: test
+  test:
+    name: Test
     strategy:
       matrix:
         include:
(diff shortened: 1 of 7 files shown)


Taowyoo:
bors r=[raoulstrackx]

Taowyoo:
Hi @raoulstrackx , could you please approve the PR instead of using bors?
The Github-Action does not works well with Github-Action, so I changed `master` branch to use GitHub Merge queue instead.

Taowyoo:
bors ping

Taowyoo:
bors merge

Taowyoo:
bors cancel

PR #313, opened 2023-09-08
Add missing setter for `conf->new_session_tickets_count`

Kazurin-775:
The file `include/mbedtls/ssl.h` in MbedTLS contains a setter for `conf->new_session_tickets_count`, which determines the number of new TLS session tickets generated for each session:

```c
#if defined(MBEDTLS_SSL_SESSION_TICKETS) && \
    defined(MBEDTLS_SSL_SRV_C) && \
    defined(MBEDTLS_SSL_PROTO_TLS1_3)
void mbedtls_ssl_conf_new_session_tickets(mbedtls_ssl_config *conf,
                                          uint16_t num_tickets);
#endif
```

However, the correspoding Rust wrapper is missing, potentially due to this function being available only when `MBEDTLS_SSL_PROTO_TLS1_3` is defined (which is not the case for the default configuration).

We should add this setter back when the `tls13` feature is enabled:

```rs
  #[cfg(feature = "tls13")]
  setter!(
      set_new_session_tickets_count(cnt: u16) = ssl_conf_new_session_tickets
  );
```

### DIFF
--- a/mbedtls/src/ssl/config.rs
+++ b/mbedtls/src/ssl/config.rs
@@ -371,6 +371,11 @@ impl Config {
         set_session_tickets(u: UseSessionTickets) = ssl_conf_session_tickets
     );
 
+    #[cfg(feature = "tls13")]
+    setter!(
+        set_new_session_tickets_count(cnt: u16) = ssl_conf_new_session_tickets
+    );
+
     setter!(set_renegotiation(u: Renegotiation) = ssl_conf_renegotiation);
 
     setter!(


Taowyoo:
Thank you for creating this PR. 
I also need to inform that: the TLS 1.3 in C mbedtls 3.X is not thread safe, so it's not recommend to use `rust-mbedtls` 0.10 and 0.11 for TLS 1.3.
Since some other user may need to just use crypto part of mbedtls, we did not yank the `rust-mbedtls` versions based on mbedtls 3.X.
Will update above in the README.

Taowyoo (approved):

PR #326, opened 2023-10-19
Introduce separate features for pki and tls functionalities of mbedtls

s-arash:
To statically guarantee that unwanted capabilities of this crate are not used (e.g., a client could only require the crypto functionality of this crate and not its ssl capabilities), this PR introduces two features: `x509` and `ssl`. `x509` enables the pki functionality of the crate and `ssl` enables its tls/ssl capabilities.

### DIFF
--- a/mbedtls/src/lib.rs
+++ b/mbedtls/src/lib.rs
@@ -15,8 +15,6 @@ compile_error!("Either the `std` or `no_std_deps` feature needs to be enabled");
 
 
 #[macro_use]
-extern crate bitflags;
-#[macro_use]
 extern crate serde_derive;
 // required explicitly to force inclusion at link time
 #[cfg(target_env = "sgx")]
@@ -37,8 +35,11 @@ pub mod hash;
 pub mod pk;
 pub mod rng;
 pub use mbedtls_platform_support::self_test as self_test;
+#[cfg(feature = "ssl")]
 pub mod ssl;
+#[cfg(feature = "x509")]
 pub mod x509;
+#[cfg(any(feature = "x509", feature = "ssl"))]
 pub mod alloc;
 
 #[cfg(feature = "pkcs12")]
--- a/mbedtls/src/lib.rs
+++ b/mbedtls/src/lib.rs
@@ -39,7 +39,7 @@ pub use mbedtls_platform_support::self_test as self_test;
 pub mod ssl;
 #[cfg(feature = "x509")]
 pub mod x509;
-#[cfg(any(feature = "x509", feature = "ssl"))]
+#[cfg(any(feature = "x509", feature = "ssl", feature = "pkcs12"))]
 pub mod alloc;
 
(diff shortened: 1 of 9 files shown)


Taowyoo:
Hi @s-arash , could you rebase your PR onto latest `main` branch, so CI will take effect.

mbedtls/src/lib.rs
```diff
@@ -37,8 +35,11 @@ pub mod hash;
 pub mod pk;
 pub mod rng;
 pub use mbedtls_platform_support::self_test as self_test;
+#[cfg(feature = "ssl")]
 pub mod ssl;
+#[cfg(feature = "x509")]
 pub mod x509;
+#[cfg(any(feature = "x509", feature = "ssl"))]
```
Taowyoo, line 42: Is it fine to restrict `alloc` under x509 or ssl ? Because this is originally a public API interface.
I guess it's better to put it under its own feature and make it as dependency for x509 or ssl.

Taowyoo (commented):
Also, regarding the version. Do you want to update version number in this PR or create another one

mbedtls/src/lib.rs
s-arash: Do you see this module being used on its own (i.e., even when `x509` and `ssl` features are not enabled)? If so, I like your suggestion, otherwise wouldn't it just add to an already long list of features that are hard to keep track of?

s-arash:
> Also, regarding the crate version. Do you want to update version number in this PR or create another one?

What's your preference? I can update the version in this PR.

mbedtls/src/lib.rs
Taowyoo: For example, `pkcs12` also depends on `alloc`.

Taowyoo: My idea is more about if external user wants to use `alloc` module, they can turn it on without turn on `x509` or `ssl` feature.
But if you stronly like current approach, we could follow it. But you need to update it to also turn on with `pkcs12`

Taowyoo:
> > Also, regarding the crate version. Do you want to update version number in this PR or create another one?
> 
> What's your preference? I can update the version in this PR.

Sure let's do it.

mbedtls/src/lib.rs
s-arash: > My idea is more about if external user wants to use alloc module, they can turn it on without turn on x509 or ssl feature.
I was thinking that this is not likely. If it's a real use-case, then adding an `alloc` feature makes sense. Since you are more familiar with the library, you should decide which approach is better.

Taowyoo (approved):

mbedtls/src/lib.rs
Taowyoo: Ok, got it.
Let just keep it simple

Taowyoo (approved):

PR #334, opened 2023-11-17
Add `Certificate::verify()` variants that return `VerifyError` as well

s-arash:
We may wish to have access to the `VerifyError` bitflags populated by the certificate verification functionality of mbedtls. This PR adds `verify()` variants that return `VerifyError` as well in case of errors.

### DIFF
--- a/mbedtls/src/x509/certificate.rs
+++ b/mbedtls/src/x509/certificate.rs
@@ -26,6 +26,8 @@ use crate::x509::{self, Crl, Time, VerifyCallback};
 #[cfg(feature = "std")]
 use yasna::{models::ObjectIdentifier, ASN1Error, ASN1ErrorKind, ASN1Result, BERDecodable, BERReader};
 
+use super::VerifyError;
+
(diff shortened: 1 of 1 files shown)


mbedtls/src/x509/certificate.rs
```diff
@@ -225,12 +227,12 @@ impl Certificate {
         err_info: Option<&mut String>,
         cb: Option<F>,
         expected_common_name: Option<&str>,
-    ) -> Result<()>
+    ) -> std::result::Result<(), (Error, VerifyError)>
```
Taowyoo, line 230: Hi @s-arash, since `rust-mbedtls` need to support no_std.
So we need to use `core::result::Result` instead of `std::result::Result`

Taowyoo (changes requested):

mbedtls/src/x509/certificate.rs
s-arash: I see! Will fix.

Taowyoo (approved):

PR #339, opened 2023-12-19
Add ciphertext check for RSADP

zugzwang:
In this PR, when decrypting with RSA without padding (e.g. `RSA_RAW_DECRYPT`), we fail early if the ciphertext is outside `{2, ..., n-2}`.

### DIFF
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -17,6 +17,7 @@ use crate::hash::Type as MdType;
 use crate::private::UnsafeFrom;
 use crate::rng::Random;
 use core::convert::TryInto;
+use core::ops::Sub;
 use core::ptr;
 
 use byteorder::{BigEndian, ByteOrder};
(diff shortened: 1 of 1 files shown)


mbedtls/src/pk/mod.rs
```diff
@@ -691,6 +692,13 @@ impl Pk {
                     return Err(Error::RsaOutputTooLarge);
                 }
 
+                // Don't process outside of {2, ..., n-2}
+                let nm1 = self.rsa_public_modulus()?.sub(&Mpi::new(1)?)?;
```
Taowyoo, line 696: So we decided to always do this FIPS check ?

zugzwang: Yes, I think it's a sane check we always want. No one should be using this API in any case, they should just use a proper RSA padding.

Taowyoo (approved):
LGTM, could you please run `cargo fmt` and push again to fix the error from fmt check?

PR #345, opened 2024-02-08
Correct `EcPoint` const time comparison and better comments and namings

Taowyoo:
- Better comments
- Better naming

### DIFF
--- a/mbedtls/src/ecp/mod.rs
+++ b/mbedtls/src/ecp/mod.rs
@@ -369,8 +369,8 @@ Please use `mul_with_rng` instead."
     ///
     /// This function will return an error if:
     ///
-    /// * `k` is not a valid private key, determined by mbedtls function [`mbedtls_ecp_check_privkey`]
-    /// * `self` is not a valid public key, determined by mbedtls function [`mbedtls_ecp_check_pubkey`]
+    /// * The scalar `k` is not valid as a private key, determined by mbedtls function [`mbedtls_ecp_check_privkey`].
+    /// * The point `self` is not valid as a public key, determined by mbedtls function [`mbedtls_ecp_check_pubkey`].
     /// * Memory allocation fails.
     /// * Any other kind of failure occurs during the execution of the underlying [`mbedtls_ecp_mul`] function.
     ///
--- a/mbedtls/src/ecp/mod.rs
+++ b/mbedtls/src/ecp/mod.rs
@@ -440,12 +440,16 @@ Please use `mul_with_rng` instead."
     /// This new implementation ensures there is no shortcut when any of `x, y ,z` fields of two points is not equal.
     ///
     /// [`mbedtls_ecp_point_cmp`]: https://github.com/fortanix/rust-mbedtls/blob/main/mbedtls-sys/vendor/library/ecp.c#L809-L825
-    pub fn eq_const_time(&self, other: &EcPoint) -> bool {
-        unsafe {
-            let x = mpi_cmp_mpi(&self.inner.X, &other.inner.X) == 0;
-            let y = mpi_cmp_mpi(&self.inner.Y, &other.inner.Y) == 0;
-            let z = mpi_cmp_mpi(&self.inner.Z, &other.inner.Z) == 0;
-            x & y & z
+    pub fn eq_const_time(&self, other: &EcPoint) -> Result<bool> {
+        let x = crate::bignum::mpi_inner_eq_const_time(&self.inner.X, &other.inner.X);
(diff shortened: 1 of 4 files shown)


Taowyoo:
Hi @zugzwang , I created this PR for solving your comments in https://github.com/fortanix/rust-mbedtls/pull/344

mbedtls/src/ecp/mod.rs
```diff
@@ -433,7 +434,12 @@ Please use `mul_with_rng` instead."
         }
     }
 
-    /// This function compares two points in const time.
+    /// This function checks equalness of two points in const time.
+    ///
+    /// The implementation is based on C mbedtls function [`mbedtls_ecp_point_cmp`].
+    /// This new implementation ensures there is no shortcut when any of `x, y ,z` fields of two points is not equal.
+    ///
+    /// [`mbedtls_ecp_point_cmp`]: https://github.com/fortanix/rust-mbedtls/blob/main/mbedtls-sys/vendor/library/ecp.c#L809-L825
```
Taowyoo, line 442: One TBD here:
The `mpi_cmp_mpi` is also not const time.
Should I use other mpi const time function to implement this?

Taowyoo: Such as this:
```c
/**
 * \brief          Check if an MPI is less than the other in constant time.
 *
 * \param X        The left-hand MPI. This must point to an initialized MPI
 *                 with the same allocated length as Y.
 * \param Y        The right-hand MPI. This must point to an initialized MPI
 *                 with the same allocated length as X.
 * \param ret      The result of the comparison:
 *                 \c 1 if \p X is less than \p Y.
 *                 \c 0 if \p X is greater than or equal to \p Y.
 *
 * \return         0 on success.
 * \return         MBEDTLS_ERR_MPI_BAD_INPUT_DATA if the allocated length of
 *                 the two input MPIs is not the same.
 */
int mbedtls_mpi_lt_mpi_ct(const mbedtls_mpi *X, const mbedtls_mpi *Y,
                          unsigned *ret);
```

mbedtls/src/ecp/mod.rs
```diff
@@ -309,10 +309,6 @@ impl EcPoint {
         Mpi::copy(&self.inner.Y)
     }
 
-    pub fn z(&self) -> Result<Mpi> {
```
mzohreva, line 312: Why is this being removed?

mbedtls/src/ecp/mod.rs
zugzwang: Yes, was going to suggest this.

mbedtls/src/ecp/mod.rs
zugzwang: Internally, Jacobian coordinates `(X, Y, Z)` are used to represent an EC point - this is purely for efficiency in addition formulæ. MbedTLS says that every function manipulating a point is expected to accept it and return it with `Z = 1` (or `Z = 0` if it is the infinity point), so that the Jacobian representation is `(x, y, 1)` where `(x, y)` are the Cartesian coordinates. Note that the previous though is only documented and not enforced in any way. Also, Jacobian representation is not unique.

mbedtls/src/ecp/mod.rs
```diff
@@ -373,9 +369,14 @@ Please use `mul_with_rng` instead."
     ///
     /// This function will return an error if:
     ///
-    /// * `k` is not a valid private key, or `self` is not a valid public key.
+    /// * `k` is not a valid private key, determined by mbedtls function [`mbedtls_ecp_check_privkey`]
```
zugzwang, line 372: I again think it does not make sense to speak about private/public keys here. It's just scalars and EC points.

But, OK, these functions is also in upstream under `ecp.c`.

For the nitpicking's sake, I suggest
```
The scalar `k` is not valid as a private key
```
and 
```
The point is not valid as a public key
```
maybe more clear.

zugzwang (approved):
LGTM but requesting changes so that we use the proper const-time compare. Thanks!

zugzwang (changes requested):
Will approve after changes addressed. Thanks for the PR

mbedtls/src/ecp/mod.rs
Taowyoo: Updated!

mbedtls/src/ecp/mod.rs
Taowyoo: Sure! Reimplemented it with calling `mbedtls_mpi_lt_mpi_ct` twice on x,y,z.

zugzwang (approved):
Thanks!

PR #347, opened 2024-02-09
Add Pk::private_from_rsa_components and Pk::public_from_rsa_components

mzohreva:


mbedtls/src/pk/mod.rs
```diff
@@ -376,6 +402,37 @@ Please use `private_from_ec_components_with_rng` instead."
         Ok(ret)
     }
 
+    pub fn private_from_rsa_components(components: RsaPrivateComponents<'_>) -> Result<Pk> {
```
Taowyoo, line 405: Nit: would be good to have some rust doc on public functions.

Taowyoo (approved):

Taowyoo:
Because of default branch name change, the CI is not triggerred now.
I will raise a PR to fix it

Taowyoo:
Hi @mzohreva , you might need to rebase this PR to trigger the CI .

Taowyoo (approved):

mbedtls/src/pk/mod.rs
```diff
@@ -138,6 +138,34 @@ const CUSTOM_PK_INFO: pk_info_t = {
...
+        /// Public exponent
+        e: &'a Mpi,
+    },
+    WithPrivateExponent {
+        /// Public modulus
+        n: &'a Mpi,
+        /// Private exponent
+        d: &'a Mpi,
+        /// Public exponent
+        e: &'a Mpi,
```
zugzwang, line 157: I understand that these values are used by upstream in `rsa_import`, but `e` is redundant here.
What is the behavior if we pass inconsistent `d, e` here (with respect to `n`)? Perhaps we should add a test, since I don't see an explicit `d, n` check in `rsa_check_context` or in `rsa_complete` which we are rightfully calling on import. It looks like it assumes that `d, n` are consistent.

zugzwang: I checked, and `rsa_check_context` does not guarantee consistency of parameters, and `rsa_complete` does not check this specifically.

Moreover I think it is an important check, for some implementations use `ed = 1 mod λ(n)` and others `ed = 1 mod φ(n)` (see e.g. https://gitlab.com/sequoia-pgp/sequoia/-/issues/792).

mbedtls/src/pk/mod.rs
```diff
@@ -1564,4 +1625,37 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
...
+        let mut pk2 = Pk::private_from_rsa_components(components).unwrap();
+        assert_eq!(pk.write_private_der_vec().unwrap(), pk2.write_private_der_vec().unwrap());
+
+        let components = RsaPrivateComponents::WithPrivateExponent {
+            n: &pk.rsa_public_modulus().unwrap(),
+            d: &pk.rsa_private_exponent().unwrap(),
+            e: &Mpi::new(pk.rsa_public_exponent().unwrap() as _).unwrap(),
+        };
+        let mut pk3 = Pk::private_from_rsa_components(components).unwrap();
+        assert_eq!(pk.write_private_der_vec().unwrap(), pk3.write_private_der_vec().unwrap());
```
zugzwang, line 1646: This test fails when I don't use the test `rng`:
```
        use crate::rng::{CtrDrbg, OsEntropy};
        let entropy = OsEntropy::new();
        let mut rng = CtrDrbg::new(entropy.into(), None).unwrap();
        let mut pk = Pk::generate_rsa(&mut rng, 2048, 0x10001).unwrap();
```

It fails in the last assertion. I think it swaps parameter order on DER serialization?

zugzwang: Perhaps we should refactor the test RNG so that it randomizes tests, or have another test RNG. (cc. @jethrogb)

mbedtls/src/pk/mod.rs
zugzwang: After further research, MbedTLS returns `RSA_BAD_INPUT_DATA` when trying to `deduce_primes`, see [here](https://github.com/fortanix/rust-mbedtls/blob/75cb9b781fa240e4ef3ba3d49cd6da5efe7f225a/mbedtls-sys/vendor/library/rsa_internal.c#L160).

This check happens a bit too late and in principle inside a loop that aborts early or late depending on the input. Arguably this is upstream but we might want to check for consistency in this struct at least. I am now wondering if this loop be used by an attacker controlling `e, n` somehow to learn information about `d`, but this is upstream concern.

mzohreva: addressed in https://github.com/fortanix/rust-mbedtls/pull/352

PR #350, opened 2024-02-12
Use fake RNG only when necessary

### LINKED ISSUE #349: Encountered test failures when randomizing tests
The current test RNG is deterministic:
```
 /// Not cryptographically secure!!! Use for testing only!!!
pub struct TestRandom(XorShiftRng);
```
and initialized in `fn test_rng()` as
```
TestRandom(XorShiftRng::new_unseeded())
```
Using a different RNG and running tests repeatedly reveals test failures:
```
    pk::tests::generate_ec_secp256r1
    pk::tests::generate_rsa
    pk::tests::private_from_rsa_components_sanity
```

(In my experiments, I replaced `pub struct TestRandom(XorShiftRng);` with `pub type TestRandom = CtrDrbg;` and initialize with `OsEntropy`)

Taowyoo:
This PR fixes #349.

- Add support of using  CtrDrbg.
- Change tests to use CtrDrbg with OsEntropy as RNG by default.

### DIFF
--- a/mbedtls/src/pk/mod.rs
+++ b/mbedtls/src/pk/mod.rs
@@ -1235,7 +1235,7 @@ iy6KC991zzvaWY/Ys+q/84Afqa+0qJKQnPuy/7F5GkVdQA/lfbhi
 
     #[test]
     fn generate_rsa() {
-        let mut pk = Pk::generate_rsa(&mut crate::test_support::rand::test_rng(), 2048, 0x10001).unwrap();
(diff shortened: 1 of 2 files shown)


zugzwang (approved):

mzohreva:
@Taowyoo if https://github.com/fortanix/rust-mbedtls/pull/351 is merged before this PR, then please use `test_rng()` in `private_from_rsa_components_sanity`, otherwise we'd need another PR for that change.

Issue #355, opened 2024-03-06
Check to set `CMAKE_TRY_COMPILE_TARGET_TYPE` too restrictive


MasterAwesome:
https://github.com/fortanix/rust-mbedtls/blob/4ff64a0b5f53d7b8b65f2a002f907852b123f85a/mbedtls-sys/build/cmake.rs#L35-L42

Targets don't necessarily end with `none-eabi` for example, if I wanted to compile for `aarch64-unknown-none` using `CFLAGS='-I <my_system_headers>' CC=clang cargo b --target aarch64-unknown-none` this fails. I think this must be configurable instead.

Removing this check, for the same command above allows me to run `mbedtls` code on baremetal aarch64.

Taowyoo:
What solution you prefer? Make it configurable through env var?

Or could you create a PR for fixing this?

PR #356, opened 2024-03-07
Allow of way to toggle freestanding platform features

MasterAwesome:
`RUST_MBED_C_COMPILER_FREESTANDING=1` allows to skip cmake checks where the target must produce valid binaries.

Closes: #355 

cc: @Taowyoo

### DIFF
--- a/mbedtls-sys/build/cmake.rs
+++ b/mbedtls-sys/build/cmake.rs
@@ -32,7 +32,9 @@ impl super::BuildConfig {
         // thumbv6m-none-eabi, thumbv7em-none-eabi, thumbv7em-none-eabihf,
         // thumbv7m-none-eabi probably use arm-none-eabi-gcc which can cause the
         // cmake compiler test to fail.
-        if target.starts_with("thumbv") && target.contains("none-eabi") {
+        if target.starts_with("thumbv") && target.contains("none-eabi")
+            || crate::features::FEATURES.have_platform_component("c_compiler", "freestanding")
+        {
             // When building on Linux, -rdynamic flag is added automatically. Changing the
             // CMAKE_SYSTEM_NAME to Generic avoids this.
             cmk.define("CMAKE_SYSTEM_NAME", "Generic");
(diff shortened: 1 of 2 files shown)


Taowyoo:
latest nightly compiler provides some new warnings. will fix them in another PR

mbedtls-sys/build/cmake.rs
```diff
@@ -32,7 +32,9 @@ impl super::BuildConfig {
         // thumbv6m-none-eabi, thumbv7em-none-eabi, thumbv7em-none-eabihf,
         // thumbv7m-none-eabi probably use arm-none-eabi-gcc which can cause the
         // cmake compiler test to fail.
-        if target.starts_with("thumbv") && target.contains("none-eabi") {
+        if target.starts_with("thumbv") && target.contains("none-eabi")
+            || crate::features::FEATURES.have_platform_component("c_compiler", "freestanding")
```
Taowyoo, line 36: This will affect the environment which already has `freestanding` feature.
I think you could just directly use read `RUST_MBED_C_COMPILER_FREESTANDING` and use it here. And it could have another name, such as `RUST_MBED_C_COMPILER_ARM_BAREMETAL` or anything more related to the issue or reason here.


In `mbedtls-sys/build/features.rs` you do not need to change the logic about adding `freestanding` feature.

Taowyoo: Also, I am thinking if your case can be determined by checking some property of `cc` here.

Note: `cc` is from `let cc = cc::Build::new().get_compiler();`

MasterAwesome: Querying from CC could work but is not guaranteed for example we can check for `-none` targets however there might be other operating systems that purely only support static libraries.

I've tested 30b7f134339c334f47563437c26fa9327a24b656 with 
```sh
$ CFLAGS="-isysroot /opt/arm-gnu-toolchain-13.2.Rel1-x86_64-aarch64-none-elf" CC=clang RUST_MBED_C_COMPILER_BAREMETAL=1 cargo +stable b --target aarch64-unknown-none --no-default-features --features no_std_deps
```

Taowyoo (approved):

PR #363, opened 2024-07-04
Yasna 0.5 and related dependencies upgrade and fmt-debug fix

mridul-manohar:
For the currently used version of yasna ver0.3, when a specially crafted DER input is provided - it will throw a panic
due to an integer overflow. The direct impact is that the executing thread aborts.
In order to overcome this vulnerability, it is advised to upgrade to latest yasna ver0.5.2.
This PR upgrades yasna version along with required dependencies.

### DIFF
--- a/mbedtls-platform-support/Cargo.toml
+++ b/mbedtls-platform-support/Cargo.toml
@@ -44,3 +44,9 @@ custom_gmtime_r = ["mbedtls-sys-auto/custom_gmtime_r", "chrono"]
 custom_time = ["mbedtls-sys-auto/custom_time", "chrono"]
 force_aesni_support = ["mbedtls-sys-auto/custom_has_support","mbedtls-sys-auto/aes_alt", "aesni"]
 aesni = ["mbedtls-sys-auto/aesni"]
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(sys_threading_component, values("custom"))',
+    'cfg(sys_time_component, values("custom"))',
+] }
--- a/mbedtls/Cargo.toml
+++ b/mbedtls/Cargo.toml
@@ -23,23 +23,23 @@ keywords = ["MbedTLS", "mbed", "TLS", "SSL", "cryptography"]
 features = ["x509", "ssl"]
 
 [dependencies]
-bitflags = "1"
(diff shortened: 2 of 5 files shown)


mbedtls/Cargo.toml
```diff
@@ -23,23 +23,23 @@ keywords = ["MbedTLS", "mbed", "TLS", "SSL", "cryptography"]
...
+bit-vec = { version = "0.6", optional = true }
 cbc = { version = "0.1.2", optional = true }
 rc2 = { version = "0.8.1", optional = true }
 cfg-if = "1.0.0"
 tokio = { version = "1.16.1", optional = true }
-chrono = { version = "0.4", optional = true }
+chrono = { version = "0.4.38", optional = true }
 
-mbedtls-sys-auto = { path = "../mbedtls-sys", version = "2.25.0", default-features = false, features = [
+mbedtls-sys-auto = { path = "../mbedtls-sys", version = "2", default-features = false, features = [
```
zugzwang, line 42: Why the downgrade?

mridul-manohar: mbedtls-sys latest version is '2.28.7' so keeping it '2' here should fetch latest one.

Taowyoo: You should use `2.28` instead of `2`

Taowyoo:
The failed CI https://github.com/fortanix/rust-mbedtls/actions/runs/10575662217/job/29380924160?pr=363 indicates there is some changes in GitHub Action's MacOS Runner.
A separate PR need to be created to fix this.

mbedtls-platform-support/Cargo.toml
```diff
@@ -27,7 +27,7 @@ chrono = { version = "0.4", optional = true }
 chrono = "0.4"
 
 [dependencies.mbedtls-sys-auto]
-version = "2.25.0"
+version = "2"
```
Taowyoo, line 30:
```suggestion
version = "2.25.0"
```
reduce version requirement is not allowed for semver.

mbedtls/Cargo.toml
Taowyoo: Please revert to "2.25.0" or change to higher.
reduce version requirement is not allowed in semver

PR #367, opened 2024-08-28
Fix and update CI

Taowyoo:
- Change to use GitHub Action to install `cargo-nextest`.
- Fix new nightly lints about `cfg`.
- Add script to run CI tests locally.


Taowyoo:
Note for reviewers:
Please ignore the CI failure `Cargo bench` because it run something on `master` which is expected broken.

mridul-manohar (approved):

PR #371, opened 2024-09-23
linking C libraries staticly and Bazel compatibility

irajtaghlidi:
This PR addresses two issues I encountered with Bazel that Cargo can handle. My approach was to solve them with minimal changes while remaining practical and aligned with Cargo and Rust guidelines.

~~* Since `mbedtls-sys` is responsible for providing the C libraries to dependent crates, we do not need to instruct Cargo to link them in the `mbedtls-platform-support` crate as well. While Cargo can handle this additional request, Bazel’s sandboxing prevents sharing the intermediate build script outputs of one crate with a dependent crate by default. Only .rlib or other library output files of dependent crates are accessible.~~

* The default linking type in `cargo:rustc-link-lib` is dynamic. However, since we are building vendor C libraries from source, we need to explicitly link them statically in the `mbedtls-sys` crate.

* Cargo has access to the entire file system at all stages, so passing `DEP_KEY=VALUE` from one crate's build script to a dependent crate's build script is possible through various methods. However, because Bazel’s Rust rules strictly limit the input data for each build task, it's preferable to share information by referencing the `OUT_DIR` of dependencies.
Now, `mbedtls-sys` crate provides `cargo:include` by pointing to its OUT_DIR, and `mbedtls-platform-support` can access header files using the `DEP_MBEDTLS_INCLUDE` environment variable. Then we don't need to customize the build process in Bazel for that.

### DIFF
--- a/mbedtls-sys/build/cmake.rs
+++ b/mbedtls-sys/build/cmake.rs
@@ -53,9 +53,9 @@ impl super::BuildConfig {
             dst.to_str().expect("link-search UTF-8 error")
         );
 
-        println!("cargo:rustc-link-lib=mbedtls");
-        println!("cargo:rustc-link-lib=mbedx509");
-        println!("cargo:rustc-link-lib=mbedcrypto");
+        println!("cargo:rustc-link-lib=static=mbedtls");
+        println!("cargo:rustc-link-lib=static=mbedx509");
+        println!("cargo:rustc-link-lib=static=mbedcrypto");
 
         println!("cargo:include={}/include", dst.to_str().expect("include/ UTF-8 error"));
         println!("cargo:config_h={}", self.config_h.to_str().expect("config.h UTF-8 error"));
--- a/mbedtls-sys/build/cmake.rs
+++ b/mbedtls-sys/build/cmake.rs
@@ -46,11 +46,10 @@ impl super::BuildConfig {
             cmk.define("CMAKE_TRY_COMPILE_TARGET_TYPE", "STATIC_LIBRARY");
         }
 
-        let mut dst = cmk.build();
+        let dst = cmk.build();
 
-        dst.push("lib");
         println!(
-            "cargo:rustc-link-search=native={}",
+            "cargo:rustc-link-search=native={}/lib",
             dst.to_str().expect("link-search UTF-8 error")
         );
 
@@ -58,14 +57,7 @@ impl super::BuildConfig {
         println!("cargo:rustc-link-lib=mbedx509");
         println!("cargo:rustc-link-lib=mbedcrypto");
 
-        println!(
-            "cargo:include={}",
-            ::std::env::current_dir()
-                .unwrap()
-                .join(&self.mbedtls_include)
-                .to_str()
-                .expect("include/ UTF-8 error")
-        );
+        println!("cargo:include={}/include", dst.to_str().expect("include/ UTF-8 error"));
         println!("cargo:config_h={}", self.config_h.to_str().expect("config.h UTF-8 error"));
     }
 }
--- a/mbedtls-sys/build/cmake.rs
+++ b/mbedtls-sys/build/cmake.rs
@@ -49,15 +49,18 @@ impl super::BuildConfig {
         let dst = cmk.build();
 
         println!(
-            "cargo:rustc-link-search=native={}/lib",
-            dst.to_str().expect("link-search UTF-8 error")
+            "cargo:rustc-link-search=native={}",
+            dst.join("lib").to_str().expect("link-search UTF-8 error")
         );
 
         println!("cargo:rustc-link-lib=static=mbedtls");
         println!("cargo:rustc-link-lib=static=mbedx509");
         println!("cargo:rustc-link-lib=static=mbedcrypto");
 
-        println!("cargo:include={}/include", dst.to_str().expect("include/ UTF-8 error"));
+        println!(
+            "cargo:include={}",
+            dst.join("include").to_str().expect("include/ UTF-8 error")
+        );
         println!("cargo:config_h={}", self.config_h.to_str().expect("config.h UTF-8 error"));
     }
 }
--- a/mbedtls-platform-support/build.rs
+++ b/mbedtls-platform-support/build.rs
@@ -34,8 +34,4 @@ fn main() {
             .flag("-ffreestanding");
     }
     b.compile("librust-mbedtls-platform-support.a");
-    // Force correct link order for mbedtls_printf
-    println!("cargo:rustc-link-lib=static=mbedtls");
-    println!("cargo:rustc-link-lib=static=mbedx509");
-    println!("cargo:rustc-link-lib=static=mbedcrypto");
 }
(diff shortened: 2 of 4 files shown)


irajtaghlidi:
Ok, Now I can see the linking error with the GNU linker (ld) which I was looking for to reproduce but since the recent version of Rust toolchain uses LLVM linker it will not happen. I'll update my PR to address it.

mbedtls-sys/build/cmake.rs
```diff
@@ -46,26 +46,18 @@ impl super::BuildConfig {
             cmk.define("CMAKE_TRY_COMPILE_TARGET_TYPE", "STATIC_LIBRARY");
         }
 
-        let mut dst = cmk.build();
+        let dst = cmk.build();
 
-        dst.push("lib");
```
jethrogb, line 51: Why? This change breaks Windows.

jethrogb:
I don't understand this PR. Are you copying the include dir to out_dir?

irajtaghlidi:
> I don't understand this PR. Are you copying the include dir to out_dir?

I don't copy. it is automatically provided in OUT_DIR after building with CMake.
```
ls "/home/iraj/fortanix/playground/bazel/rust-mbedtls/target/debug/build/mbedtls-sys-auto-6b1c4dec2d5df1f4/out"
bindings.rs  build  config.h  include  lib  mod-bindings.rs
```

jethrogb:
> it is automatically provided in OUT_DIR after building with CMake.

Oh, interesting.

mbedtls-sys/build/cmake.rs
irajtaghlidi: I was using the `dst` in a second place, so removed the `push()`. Now it is fixed by using `join()` instead. Is it compatible with Windows? Since tests passed I couldn't validate this part.

irajtaghlidi:
I had to revert my first change to solve linking issue with `mbedtls_printf` method in `mbedtls-platform-support` lib crate. My effort was to make `rust-mbedtls` builds natively with Bazel without any extra customization. But it seems both `mbedtls-sys` and `mbedtls-plaform-support` crates need to be linked to C static libraries.

For the reference:
Then we have to provide `mbedtls-sys`'s build script output dir as a data input for the `mbedtls-platform-support` using annotation in `WORKSPACE.bazel` like this example:

```
crates_repository(
    # ...
    annotations = {
        "mbedtls-platform-support": [crate.annotation(
            # compile_data or data
            data = [
                "@crate_index__mbedtls-sys-auto-2.28.7//:build_script_build",
            ],
        )],
    }
}
```

mbedtls-sys/build/cmake.rs
```diff
@@ -46,25 +46,20 @@ impl super::BuildConfig {
...
-        println!("cargo:rustc-link-lib=mbedcrypto");
+        println!("cargo:rustc-link-lib=static=mbedtls");
+        println!("cargo:rustc-link-lib=static=mbedx509");
+        println!("cargo:rustc-link-lib=static=mbedcrypto");
 
         println!(
             "cargo:include={}",
-            ::std::env::current_dir()
-                .unwrap()
-                .join(&self.mbedtls_include)
```
jethrogb, line 65: Not sure if this var is used anywhere else.

irajtaghlidi: It doesn't. This part is just trying to provide header files location to dependent crates that can be read by `DEP_MBEDTLS_INCLUDE` environment variable. Either from the source directory or alternatives (out_dir). and from dependent crates, it does not matter as long as it has read access.
Cargo uses a flat model and you have full access to the file system but Bazel sandboxing requires some practices to follow for more compatibility.

jethrogb: So then the var can be deleted?

irajtaghlidi: `self.mbedtls_include` is still being used by `Bindgen` but I'll check if headers are also accessible in there, Since it's running after the CMake step.
```
fn main() {
    let cfg = BuildConfig::new();
    cfg.create_config_h();
    cfg.print_rerun_files();
    cfg.cmake();
    cfg.bindgen();
}
```

irajtaghlidi: Now, Only CMake depends on the source path and other components are using OUT_DIR content.

mbedtls-sys/build/cmake.rs
```diff
@@ -46,25 +46,20 @@ impl super::BuildConfig {
...
-            dst.to_str().expect("link-search UTF-8 error")
+            dst.join("lib").to_str().expect("link-search UTF-8 error")
         );
 
-        println!("cargo:rustc-link-lib=mbedtls");
-        println!("cargo:rustc-link-lib=mbedx509");
-        println!("cargo:rustc-link-lib=mbedcrypto");
+        println!("cargo:rustc-link-lib=static=mbedtls");
+        println!("cargo:rustc-link-lib=static=mbedx509");
+        println!("cargo:rustc-link-lib=static=mbedcrypto");
```
Taowyoo, line 58: What's the affect of newly added `static=`? 
Does it just explicitly ensure static link?

irajtaghlidi: > What's the affect of newly added `static=`?
> Does it just explicitly ensure static link?

It does not have an effect in Cargo but setting static linking explicitly makes Bazel's Rust rule work properly.

More Context:
Since they are built script's output in a lib (rlib) crate, it should and would linked statically to the final ELF and Cargo handles it automatically even if we don't mention it.
But Rust's rule in Bazel is trusting what we asked for. Since the default value for linking is dynamic. it tries to find these libraries system defined shared libraries and causes failure.
[link the generated crate to a native library](https://doc.rust-lang.org/rustc/command-line-arguments.html#-l-link-the-generated-crate-to-a-native-library)

So it would be better for both to define explicitly.

irajtaghlidi:
Hi team, please review this PR and share your final thoughts.

mbedtls-sys/build/cmake.rs
Pagten: @irajtaghlidi why doesn't this change have an effect when building with Cargo?

On the page you linked above, I see

> The kind of library and the modifiers can also be specified in a [#[link] attribute](https://doc.rust-lang.org/reference/items/external-blocks.html#the-link-attribute). If the kind is not specified in the link attribute or on the command-line, it will link a dynamic library by default, except when building a static executable. If the kind is specified on the command-line, it will override the kind specified in a link attribute.

In the original code, the kind is not specified in the `link` attribute nor on the command line, so wouldn't cargo try to link a dynamic library?

I understand that the mbedtls-sys build script builds a static lib version of mbedtls and hence we want to link that static lib, but I haven't spotted how the build script tells Cargo that it's a static lib. The documentation (or at least my interpretation of it) seems to indicate that Cargo would try to link a dynamic lib by default, so I would expect the original code to fail. Probably I'm missing something...

jethrogb: My guess is Rust tries to link in whatever way it can, it finds that the only available option is static, and does that.

I also note that `mbedtls-platform-support` already does this:
```rust
println!("cargo:rustc-link-lib=static=mbedtls");
println!("cargo:rustc-link-lib=static=mbedx509");
println!("cargo:rustc-link-lib=static=mbedcrypto");
```

jethrogb: Anyway, this specific change seems fine.

Pagten (approved):

PR #376, opened 2025-01-09
Set thread names in tests

jethrogb:


mbedtls/tests/support/mod.rs
```diff
@@ -7,6 +7,7 @@
  * according to those terms. */
 
 #![allow(dead_code)]
+
```
Taowyoo, line 10:
```suggestion
```
nit: extra empty line here

Taowyoo (approved):

PR #396, opened 2025-05-14
Js/mbedtls 0.12.5 merge

janosimas:
Branch created from the latest commit before v0.13 and merge from v0.12.4


Taowyoo:
I guess one of #395 and #396 is duplicated ?

PR #410, opened 2026-06-23
Reduce amount of pointer loads in RSA hot loops

zugzwang:
LVI mitigations apply LFENCE instructions whenever a pointer is loaded. In the context of Montgomery multiplication for RSA, this degrades performance of private operations. This patch prepares loaded pointers before hot loops in order to reduce the degradation.

### DIFF
--- a/mbedtls-sys/Cargo.toml
+++ b/mbedtls-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mbedtls-sys-auto"
-version = "2.28.14"
+version = "2.28.15"
 authors = ["Jethro Beekman <jethro@fortanix.com>"]
 build = "build/build.rs"
 license = "Apache-2.0 OR GPL-2.0-or-later"
--- a/mbedtls-sys/vendor/library/bignum.c
+++ b/mbedtls-sys/vendor/library/bignum.c
@@ -142,8 +142,9 @@ int mbedtls_mpi_shrink(mbedtls_mpi *X, size_t nblimbs)
     }
     /* After this point, then X->n > nblimbs and in particular X->n > 0. */
 
+    const mbedtls_mpi_uint *Xp = X->p;
     for (i = X->n - 1; i > 0; i--) {
-        if (X->p[i] != 0) {
+        if (Xp[i] != 0) {
             break;
         }
     }
@@ -157,7 +158,7 @@ int mbedtls_mpi_shrink(mbedtls_mpi *X, size_t nblimbs)
         return MBEDTLS_ERR_MPI_ALLOC_FAILED;
     }
 
-    if (X->p != NULL) {
+    if (Xp != NULL) {
         memcpy(p, X->p, i * ciL);
         mbedtls_mpi_zeroize(X->p, X->n);
         mbedtls_free(X->p);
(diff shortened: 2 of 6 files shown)


mbedtls-sys/vendor/library/bignum.c
```diff
@@ -1880,8 +1901,11 @@ int mbedtls_mpi_mod_int(mbedtls_mpi_uint *r, const mbedtls_mpi *A, mbedtls_mpi_s
     /*
      * general case
      */
-    for (i = A->n, y = 0; i > 0; i--) {
-        x  = A->p[i - 1];
+
+    const size_t An = A->n;
```
s-arash: `An` looks unnecessary.

mbedtls-sys/vendor/library/bignum.c
```diff
@@ -1927,6 +1951,8 @@ void mbedtls_mpi_montmul(mbedtls_mpi *A,
                          mbedtls_mpi_uint mm,
                          const mbedtls_mpi *T)
 {
+    // This is just a marker so that we know the patch is used
+    asm volatile("" :: "r"(mbedtls_montmul_patch_marker) : "memory");
```
s-arash: Do we want to keep this?

mbedtls-sys/vendor/library/bignum.c
zugzwang: Right, and also it's only a `size_t`

mbedtls-sys/vendor/library/bignum.c
zugzwang: It doesn't hurt in any way, and I think it's handy to check the patch is applied

jethrogb: I don't really understand the point of this. Are we worried that we are compiling code from the wrong source tree? Why do we have that worry only for this change, and not for _every_ other code change we do?

mbedtls-platform-support/Cargo.toml
Taowyoo, line 3: we need to use 0.1.4 if  need to publish new one
ref: https://crates.io/crates/mbedtls-platform-support/versions

mbedtls-sys/Cargo.toml
Taowyoo: Similar to my previous comment, you may need to use 2.28.15

mbedtls/Cargo.toml
Taowyoo, line 5: similarly, 0.13.6

jethrogb:
I feel like a lot of unrelated changes are being combined in this single PR?

mbedtls-sys/vendor/library/bignum.c
zugzwang: Yes, I was using this as a way to quickly check the patch (version), as I was juggling with several rust binaries for performance testing. But, fair point, we won't be needing it anymore.

zugzwang:
@jethrogb 

> I feel like a lot of unrelated changes are being combined in this single PR?

I retargeted and cleaned this, there were a couple of commits that slipped in from a release branch. Current changes are only related to the patch. And I also removed the assembly marker as per your comment.

Taowyoo (approved):

Taowyoo (approved):

PR #412, opened 2026-08-13
Fix GitHub Actions workflow for crate publishing

Taowyoo:
This PR:

1. Fix the crate publish job.
2. Add a dry-run step before publishing.


Taowyoo:
Action is validated by https://github.com/fortanix/rust-mbedtls/actions/runs/31753772331
The dry run succeeds.
Then I reject it, so it shows Failure

PR #413, opened 2026-09-08
Add mitigation against CVE-2025-66442

anilburakgokce:
This PR makes sure the necessary mitigations are employed against the vulnerability published [here](https://mbed-tls.readthedocs.io/en/latest/security-advisories/mbedtls-security-advisory-2026-03-compiler-induced-constant-time-violations/).

### DIFF
--- a/mbedtls-sys/build/cmake.rs
+++ b/mbedtls-sys/build/cmake.rs
@@ -31,8 +31,10 @@ impl super::BuildConfig {
             cmk.cflag(cflag);
         }
         let cc = cc::Build::new().get_compiler();
+        let target = std::env::var("TARGET").expect("TARGET environment variable should be set in build scripts");
         if cc.is_like_clang() && cc.args().iter().any(|arg| arg == "-mllvm") {
             cmk.define("CMAKE_C_COMPILER_FORCED", "TRUE");
+            mitigate_cve_2022_66442(&target, &mut cmk);
         }
 
         println!("cargo:rerun-if-env-changed=RUST_MBED_C_COMPILER_BAREMETAL");
@@ -40,7 +42,6 @@ impl super::BuildConfig {
             .map(|val| val == "1")
             .unwrap_or_default();
 
-        let target = std::env::var("TARGET").expect("TARGET environment variable should be set in build scripts");
         // thumbv6m-none-eabi, thumbv7em-none-eabi, thumbv7em-none-eabihf,
         // thumbv7m-none-eabi probably use arm-none-eabi-gcc which can cause the
         // cmake compiler test to fail.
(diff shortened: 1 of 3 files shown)


mbedtls-sys/build/cmake.rs
```diff
@@ -74,3 +75,21 @@ impl super::BuildConfig {
...
     }
 }
+
+use super::config;
+
+/// To mitigate CVE-2025-66442, make sure select-optimize is disabled when necessary
+/// returns true iff "-mllvm" and "--disable-select-optimize=true" flags are added to cmk
+/// Note: this function is public only because we have a unit test for it in test/cmake_tests.rs
+/// Note: this function returns a boolean so that we can confirm its functionality via a unit test
+pub fn mitigate_cve_2022_66442(cc_is_like_clang: bool, target: &String, cmk: &mut cmake::Config) -> bool {
```
zugzwang: Nit: 2025

mbedtls-sys/build/cmake.rs
```diff
@@ -74,3 +75,21 @@ impl super::BuildConfig {
...
+    let target_asm_protected =
+        target.contains("x86_64") || target.contains("i686") || target.starts_with("arm") || target.starts_with("aarch64");
+    let mbedtls_have_asm = config::default_defines().get("MBEDTLS_HAVE_ASM") == Some(&config::Macro::Defined);
```
zugzwang: If I get this right, `default_defines` is the default, and it does have `MBEDTLS_HAVE_ASM`. So this will always be `true`, regardless of the actual flags used.

anilburakgokce: You are right, but I wanted to make sure that a future change in this behavior (`MBEDTLS_HAVE_ASM` being always `true`) does not cause a break in this mitigation.

mbedtls-sys/build/cmake.rs
anilburakgokce: Nice catch, thanks!

mbedtls-sys/build/cmake.rs
zugzwang: I see, so when someone in the future goes and `Undefine`s the flag in the default config.

Any ideas for checking flags for the current build (not default)?

anilburakgokce: It is rather complicated but possible. The features are stored in a HashMap (`defines`) in the `create_config_h` method, but that method creates the `config.h` file and disposes the hash map.

Option 1: Once the `defines` is constructed, we can search for `MBEDTLS_HAVE_ASM` in it, and store the result in a new field of `self`.
Option 2: In `mitigate_cve_2025_66442`, we can scan the whole `config.h` file to try to find the line `#define MBEDTLS_HAVE_ASM`.

I am closer to Option 1. Shall I do it?

Edit: I implemented Option 1, can you please review?

mbedtls-sys/build/build.rs
```diff
@@ -30,10 +31,11 @@ struct BuildConfig {
     mbedtls_src: PathBuf,
     config_h: PathBuf,
     cflags: Vec<String>,
+    mbedtls_have_asm_defined: bool,
```
zugzwang: Could you add a comment stating that this is in order to mitigate CVE-2025-66442?

zugzwang (approved):
Thanks for the PR!

